PR #7, opened 2016-07-23
Clear up installation instructions

dawsonbotsford:
The current method leaves questions for how this works for new developers especially


keithamus:
Just FYI, you can install the cli tool without `--global`, if you have `./node_modules/.bin` in your `PATH`

PR #16, opened 2017-06-08
Add `module`, which is the new `jsnext:main`

swansontec:
Both rollup.js and Webpack 2 now support `module` over `jsnext:main`,
so let's get that added.

See the [Rollup.js docs](https://github.com/rollup/rollup/wiki/pkg.module).

### DIFF
--- a/index.js
+++ b/index.js
@@ -78,6 +78,7 @@ function sortPackageJson(packageJson) {
     'contributors',
     'files',
     'main',
+    'module',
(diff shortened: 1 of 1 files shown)


keithamus:
Great work!

keithamus:
Should be released as `1.7.0` within the next 10 minutes hopefully 😄

swansontec:
Thanks, awesome!

PR #23, opened 2018-02-03
array_unique keywords

bluelovers:
array_unique keywords


keithamus:
What is the motiviation behind this change? I don't understand why this is needed.

bluelovers:
@keithamus sory i thinking u are not going look this so i push it to npm
i will re-pr again

PR #24, opened 2018-02-04
array_unique keywords 2

bluelovers:


keithamus:
Hey @bluelovers this is still missing any rationale. Why do we need to make items unique? Surely this could lose data while running? Not every array in a package.json is going to be unique.

bluelovers:
keywords is string array, and it is easy has duplicate value
no reason about need duplicate value

keithamus:
Ah yes I see. Happy to merge then 😄

PR #25, opened 2018-02-22
Add Jest as a sortable key

selbekk:
This commit adds `jest` as a sortable key, right after the linting keys.

[Jest offers tons of configuration](https://facebook.github.io/jest/docs/en/configuration.html) that you might want to stick in your `package.json`. I came across this project because I needed it, and this key was missing based on our own needs. Figuring since Jest is at least one of the most popular testing frameworks out there, this should make the cut.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,6 +61,7 @@ function sortPackageJson(packageJson) {
   sortSubKey('browserify');
   sortSubKey('babel');
   sortSubKey('eslintConfig');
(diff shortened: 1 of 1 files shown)


selbekk:
By the way - really cool project that does what it says on the tin - thanks a ton bud!

keithamus:
Good work @selbekk 👍. happy this is of use to you. Should be released as `1.8.0`

PR #26, opened 2018-02-22
feat: Add support for specifying location

selbekk:
This commit adds support for specifying location(s) for the `package.json` file(s) to sort. 

This is particularly useful when you have a monorepo with tons of `package.json` files that need this command.

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,13 +124,18 @@ function sortPackageJson(packageJson) {
 }
 module.exports = sortPackageJson;
 module.exports.sortPackageJson = sortPackageJson;
+
 if (require.main === module) {
   var fs = require('fs');
-  var packageJsonPath = process.cwd() + '/package.json';
-  var packageJson = fs.readFileSync(packageJsonPath, 'utf8');
-  var sorted = sortPackageJson(packageJson);
-  if (sorted !== packageJson) {
-    fs.writeFileSync(packageJsonPath, sorted, 'utf8');
-    console.log('Ok, your package.json is sorted');
-  }
+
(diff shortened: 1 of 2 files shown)


selbekk:
Thought I'd pass this as well, this was basically just me hacking away on your code to make it work for our particular use case (we have a lerna monorepo with ~ 20 packages, and I'm really bad at bash-scripting. 🙈 

Hopefully it'll be helpful for others as well!

I'm sure there are better ways to solve this than what I did, so if you want me to tweak stuff or w/e, please let me know =)

index.js
```diff
@@ -124,13 +124,19 @@ function sortPackageJson(packageJson) {
...
   if (sorted !== packageJson) {
-    fs.writeFileSync(packageJsonPath, sorted, 'utf8');
-    console.log('Ok, your package.json is sorted');
+      fs.writeFileSync(filePath, sorted, 'utf8');
+      console.log(filePath + ' is sorted!');
   }
 }
+
+if (require.main === module) {
+  var fs = require('fs');
```
keithamus: I don't like that we require fs here, and up above on line L132 use fs. This is relying on lexical scope hoisting which - although it works - I dislike. Perhaps just make `processFile` an anonymous function to `forEach`?

Otherwise this code looks 👍.

selbekk: Totally agree - this was a result of refactoring without looking at the whole picture.

selbekk:
yaaaaas

keithamus:
`1.9.0` 🎉. Good work @selbekk! And thanks for sticking to commit conventions!

PR #27, opened 2018-02-23
Add support for `prettier` key

simonkberg:
Prettier is a popular code formatter. It has over 21k stars on GitHub and 3M monthly downloads on npm.

It supports configuration via the "prettier" key in `package.json`, so I think it makes sense to add support for it here. https://prettier.io/docs/en/configuration.html

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,6 +62,7 @@ function sortPackageJson(packageJson) {
   sortSubKey('babel');
   sortSubKey('eslintConfig');
   sortSubKey('jest');
+  sortSubKey('prettier');
   sortSubKey('dependencies');
   sortSubKey('devDependencies');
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -109,6 +110,7 @@ function sortPackageJson(packageJson) {
     'eslintConfig',
     'stylelint',
     'jest',
+    'prettier',
```
keithamus: I know this is super picky, but I think we should bump this to just below `eslintConfig` - or maybe above. That way all linting tools config goes first, then test config, then dependencies.

simonkberg: Makes sense, updated

keithamus:
Will be `1.11.0`

PR #34, opened 2018-04-21
feat: add support for maintaining `package.json` indentation level when sorting

### LINKED ISSUE #33: Maintain indenting level
NPM respects the indentation level of `package.json`; if you use 4 spaces for indenting, npm will maintain that, rather than reverting it back to 2 spaces everytime it makes a change to the `package.json`.

Currently it seems `sort-package-json` doesn't respect indentation level, and just instead reverts the `package.json` to 2 spaces.

G-Rath:
This is the same way `npm` does it.

I had to alter a test that was actually testing specifically that the indentation had been formatted, so now it's testing that the indentation hasn't changed.

fixes #33

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,13 @@
 #!/usr/bin/env node
 var sortObjectKeys = require('sort-object-keys');
+var detectIndent = require('detect-indent');
 function sortPackageJson(packageJson) {
   var wasString = false;
   var endCharacters = '';
+  var indentLevel = 2;
   if (typeof packageJson === 'string') {
     wasString = true;
+    indentLevel = detectIndent(packageJson).indent;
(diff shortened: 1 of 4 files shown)


G-Rath:
This also generated a `package-lock.json`, that I've not committed since one doesn't already exist, but I would advise adding one. 

Let me know if you'd like me to add it as part of this PR.

keithamus:
This looks awesome @G-Rath! Please do go ahead and add the package-lock.json, also if you could add one or two tests just to test the indenting behavior, so we know if it were to fail, that'd be awesome!

G-Rath:
@keithamus done and done 👍

keithamus:
Wonderful work @G-Rath! Should be released very soon 😉

PR #35, opened 2018-05-05
tweak order

bluelovers:


keithamus:
👍  thanks @bluelovers - both PRs will be released shortly 😃

PR #44, opened 2018-10-09
Add VSCode extension mention

unional:
Hi, I have created a VSCode extension using your library.

Adding a link in the README so that people can also use that in VSCode. 🌷


keithamus:
Awesome!

PR #50, opened 2018-12-02
feat: Keep the "pre" when sorting certain scripts

Shingyx:
Before, "prepare" was being sorted as if it was named "pare" and "prettier" was being sorted as if it was named "ttier". These scripts are now sorted as is. This is done via a blacklist, so other similar scripts could be added later.

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,6 +56,10 @@ var sortOrder = [
   'preferGlobal',
   'publishConfig',
 ];
+var prefixedScriptBlacklist = [
+  'prepare',
+  'prettier',
+];
 
 function sortPackageJson(packageJson) {
   var wasString = false;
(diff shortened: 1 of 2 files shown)


keithamus:
@Shingyx thanks for the PR! I think I'd prefer a whitelist instead, as this keeps coming up. Perhaps we could whitelist all of the known `pre` and `post` npm hooks and only sort those before?

Thanks!

Shingyx:
Fair enough. I'll take the list from https://docs.npmjs.com/misc/scripts, plus any scripts in the package.json that don't start with `pre` or `post`. How does that sound?

index.js
```diff
@@ -69,6 +83,18 @@ function sortPackageJson(packageJson) {
...
     packageJson = JSON.parse(packageJson);
   }
+
+  var prefixedScriptRegex = /^(pre|post)(.)/;
+  var prefixableScripts = defaultNpmScripts.slice();
+  if (typeof packageJson.scripts === 'object') {
+    Object.keys(packageJson.scripts).forEach(function (script) {
+      var prefixOmitted = script.replace(prefixedScriptRegex, '$2');
+      if (packageJson.scripts[prefixOmitted] && prefixableScripts.indexOf(prefixOmitted) < 0) {
+        prefixableScripts.push(prefixOmitted);
```
keithamus: This is great stuff! A nice piece of logic here!

PR #54, opened 2019-01-27
chore(index.js): upgrade source and tests to es6

amilajack:


keithamus:
Hey @amilajack thanks for the PR!

It looks like this does a little more than just update to ES6. There looks to be some changes to functionality here also. Could you explain those please?

Thanks

amilajack:
@keithamus the diff looked weird because there were some changes made to master. The PR should be easier to review now.

PR #55, opened 2019-02-03
feat(index): allow custom ordering of package.json properties

amilajack:


keithamus:
Great work 👍! All for making this easier to consume as a library 😄

PR #57, opened 2019-03-01
feat: add several bundle entries (umd:main, unpkg, source), and flat …

gillesdandrea:
Add several bundle entries: `umd:main`, `unpkg`, `source`
Add `flat` used by Yarn.

See https://github.com/stereobooster/package.json


keithamus:
🎉 thanks @gillesdandrea!

PR #60, opened 2019-03-25
publish TypeScript typings in "files" attribute and reference them in package.json "typings" attribute

vincentbriglia:


keithamus:
👍

Issue #67, opened 2019-07-05
--lint / --check mode


kachkaev:
This is a great tool, I'm really glad that I've found it! 🙌

As a maintainer of a monorepo, I would like to use `sort-package-json` in PR pipelines to prevent unsorted `package.json` files from landing the `master` branch. Basically I'd like the tool to become a part of the linting task alongside `eslint`,  `tsc` and `prettier`. Would it be possible to add a new option called `--lint` or `--check`, which instead of writing the files would make `sort-package-json` just return a non-zero exit code if there is a diff?

I know that I can already set a pre-commit hook, but it can be skipped by a contributor. It is also possible to run `sort-package-json` in CI ‘normally’ and then check the git diff, but that's a bit laborious. Here's what looks to me as an decent solution (note `lint:package-json`):

```json
{
  "scripts": {
    "fix": "yarn fix:package-json && yarn fix:prettier",
    "fix:package-json": "sort-package-json package.json packages/*/package.json",
    "fix:prettier": "prettier --write \"**/*.*\"",
    "lint": "yarn lint:package-json && yarn lint:prettier && lint:tsc",
    "lint:package-json": "sort-package-json --check package.json packages/*/package.json",
    "lint:prettier": "prettier --check \"**/*.*\"",
    "lint:tsc": "yarn workspaces run lint:tsc",
  }
}
```

See [Prettier's `--check`](https://prettier.io/docs/en/cli.html#check) as an inspiration.

dosentmatter:
If you are up for it, you can just use the [API](https://github.com/keithamus/sort-package-json/blob/de04f882fc88434f4895788e22ed63531a05c053/index.js#L200-L202).

It only overwrites files if you [run it on the CLI](https://github.com/keithamus/sort-package-json/blob/de04f882fc88434f4895788e22ed63531a05c053/index.js#L204-L218).

From reading the code, you can pass either a string or object to `sortPackageJson()` and it will return back the same type as you gave it. On the CLI, the code depends on [string comparison](https://github.com/keithamus/sort-package-json/blob/de04f882fc88434f4895788e22ed63531a05c053/index.js#L214) to check if sorting was done.

If you pass `sortPackageJson()` an object, it will return a new object, so you can't really use that to check if sorting was done, since they are different references.

[...]

PR #86, opened 2019-12-02
feat: support `--check` flag

fisker:
Add a simple implantation of `--check` flag.

Notice: no tests added.

fixes: #67


fisker:
tests added, should be good enough for `--check`. maybe we can improve later

Issue #84, opened 2019-12-01
Sorting of "exports" field


RebeccaStevens:
Conditional Exports:
https://nodejs.org/api/esm.html#esm_conditional_exports

Related to #76

Personally I would like to have this be sorted directly after "main" as it essentially replaces fields like "module", "browser" etc.

keithamus:
Sounds like a good idea. PRs welcome!

fisker:
@RebeccaStevens check this comment https://github.com/keithamus/sort-package-json/pull/88#issuecomment-560997003

PR #88, opened 2019-12-03
feat: add `exports` field

fisker:
https://nodejs.org/api/esm.html#esm_package_exports

fixes: #84

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,7 @@ const sortOrder = [
   'files',
   'sideEffects',
   'type',
+  'exports',
   'main',
   'umd:main',
   'jsdelivr',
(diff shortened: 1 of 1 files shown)


fisker:
I didn't see #84 when I send this PR, I think better put `exports` above `main`, since this could be object.

If we put it under `main`, it should go under all those fields

```text
  'main',
  'umd:main',
  'jsdelivr',
  'unpkg',
  'module',
  'source',
  'jsnext:main',
  'browser',
  'types',
  'typings',
  'style',
  'example',
  'examplestyle',
  'assets',
  'bin',
  'man',
```

Issue #83, opened 2019-12-01
Opt-out of sorting keywords


robatwilliams:
The order of `keywords` is preserved when displayed on npmjs.com.

Packages may wish to order keywords such that related ones are together. For example: what it does, what tech it's for, what tech it uses, other related packages.

keithamus:
You raise a good point. Let's just unilaterally disable the sorting of keywords. PRs welcome!

niksy:
Sorry for commenting on closed issue, but is there a possibility to sort keywords alphabetically, or some sort function we can use to modify sorting method just for keywords?

keithamus:
Hey @niksy! You can create a fork of this repository and change it, perhaps publish it as `sort-package-json-and-keywords`?

niksy:
@keithamus thanks!

PR #90, opened 2019-12-05
refactor: major code refactor

fisker:
Refactor sortSubKey logic, prepare for #83 to ignore some field.

I'm planning to move all this to sortOrder, only problem is `compareScriptKeys` is in the inner scope, will try something else later


keithamus:
To fix #83 we can just avoid sorting `keywords`. Is there another justification for this refactor?

fisker:
I'm trying to make it easier to add new field,

we don't need add `mocha` to `sortOrder`, then add a new line `sortSubKey('mocha')`,  we can add `{field: 'mocha', sort: true}` something like this

fisker:
double checked, I didn't miss any field

keithamus:
I think it might be useful to minimise. Let's see a complete implementation then, that combines `sortOrder` with these subkeys.

fisker:
I look at #83 again,  only ignore `keywords` from sorting, might don't need this.
I thought it ask for ignore some fields. 

But I think still useful to put all this together.

One question, should we support shortcut `"key"` for `{key, sort: true}`(28 of them) or `{key, sort: false}`(31 of them)

index.js
```diff
@@ -119,15 +130,31 @@ function sortPackageJson(packageJson, options = {}) {
...
       return;
     }
+
     if (typeof packageJson[key] === 'object') {
+      let sortList
+      if (sortScripts) {
+        sortList = compareScriptKeys
+      } else if (Array.isArray(sort)) {
+        sortList = sort
+      }
```
keithamus: This logic seems a little complicated. Can we just provide `sortList` as an option to the function?

index.js
```diff
@@ -3,76 +3,84 @@ const sortObjectKeys = require('sort-object-keys');
...
+  {key: 'version', sortSubKey: false},
+  {key: 'private', sortSubKey: true},
+  {key: 'description', sortSubKey: false},
+  {key: 'keywords', sortSubKey: true, unique: true},
+  {key: 'homepage', sortSubKey: true},
+  {key: 'bugs', sortSubKey: true, sortList: ['url', 'email']},
+  {key: 'repository', sortSubKey: true, sortList: ['type', 'url']},
+  {key: 'funding', sortSubKey: true, sortList: ['type', 'url']},
+  {key: 'license', sortSubKey: true, sortList: ['type', 'url']},
+  {key: 'author', sortSubKey: true, sortList: ['name', 'email', 'url']},
```
fisker: Do you think this is a little complicated?

I was trying to make this simplier like this


```js
{
  'name',
  'version',
  {key: 'homepage'},
  {key: 'bugs', sortList: ['url', 'email']}
}
```

But, seems not very clear that `homepage` and `bugs` will sort sub key, any suggestion?

keithamus:
@fisker I've done a bit more refactoring, cleaned up a lot of stuff. Let me know if you think this is good?

index.js
```diff
@@ -85,183 +24,245 @@ const defaultNpmScripts = [
...
+  { key: 'private' },
+  { key: 'description' },
+  { key: 'keywords', over: uniq },
+  { key: 'homepage' },
+  { key: 'bugs', over: sortObjectBy(['url', 'email']) },
+  { key: 'repository', over: sortURLObject },
+  { key: 'funding', over: sortURLObject },
+  { key: 'license', over: sortURLObject },
+  { key: 'author', over: sortAuthorObject },
+  { key: 'files' },
```
fisker: I think this was sorted before, as array, maybe we can also unique it

fisker:
Looks great,  I'll check again when I get back to my laptop

package.json
```diff
@@ -35,14 +36,35 @@
       "commit-msg": "validate-commit-msg"
     }
   },
+  "prettier": {
+    "semi": false,
+    "singleQuote": true,
+    "trailingComma": "all"
```
fisker:
```suggestion
    "trailingComma": "es5"
```

index.js
fisker: sorry, it's not sorted before,  in this case

```js
  "files": [
    "index.js",
    "index.d.ts"
  ],
```

I guess , it doesn't make sense to put `index.d.ts` above `index.js`, wo we should keep the order as it is.

add uniq in https://github.com/keithamus/sort-package-json/pull/90/commits/e10cf9ad3389d13e27a0e617419718db6a7ae78a

fisker:
@keithamus I did a little change, let me know if there is any problem

fisker:
You did great  job on this refactoring, like to see it get merged.

index.js
```diff
@@ -85,183 +27,209 @@ const defaultNpmScripts = [
...
-  let hasWindowsNewlines = false;
-  let endCharacters = '';
-  let indentLevel = 2;
-  if (typeof packageJson === 'string') {
-    wasString = true;
-    indentLevel = detectIndent(packageJson).indent;
-    if (packageJson.substr(-1) === '\n') {
-      endCharacters = '\n';
+const sortScripts = scripts => {
+  const prefixableScripts = defaultNpmScripts
```
fisker: @keithamus this should be a copy of `defaultNpmScripts` https://github.com/keithamus/sort-package-json/blob/bade9646d07d0c79059815701f2fdfde3e0e2f82/index.js#L109

fisker:
Good to go.

PR #94, opened 2019-12-06
fix: sort `betterScripts` in same order of `scripts`

fisker:
`betterScripts` should sort just like `scripts`


keithamus:
@fisker I have the same kind of change that I've pushed to your PR here: https://github.com/keithamus/sort-package-json/pull/90

fisker:
close in favor of #90

PR #98, opened 2019-12-16
refactor: improve cli

fisker:
1. switch `glob` to `globby`, `globby` is easier to use, and `fast-glob`(globby uses) is faster on large projects.
2. improve `--check` messages, add matched files count
3. add a message when not matching any file


fisker:
@keithamus done.

PR #101, opened 2019-12-20
refactor: separate cli from `index.js`

fisker:
Use a separate cli.js, should be easier to maintain.


fisker:
@keithamus rebased.

Issue #20, opened 2017-12-12
feature request: group git hooks together


reggi:
Here's a list of git hooks that can be applied to the npm scripts. Thoughts?

https://github.com/typicode/husky/blob/HEAD/HOOKS.md

keithamus:
Seems totally reasonable to group these together! Happy to see a PR that adds this feature 😄

muuvmuuv:
@reggi link no longer working and you going to take on this?

reggi:
@muuvmuuv unable to work on this

fisker:
Can we close this now?

PR #103, opened 2019-12-20
feat: sort husky hooks

fisker:
Not sure is this #20 requested, we should sort them anyway.


index.js
```diff
@@ -15,6 +15,38 @@ const sortObject = onObject(sortObjectKeys())
...
+    'post-update',
+    'push-to-checkout',
+    'pre-auto-gc',
+    'post-rewrite',
+    'rebase',
+    'sendemail-validate',
+    'fsmonitor-watchman',
+    'p4-pre-submit',
+    'post-index-change',
+  ],
```
fisker: You prefer keep this list here, or a separate package I build https://github.com/fisker/git-hooks-list ?

keithamus: Yes I was thinking we could start moving a lot of this stuff out to dependencies. I'm happy to add this dependency 👍

fisker: done.

PR #104, opened 2019-12-20
fix: remove `betterScripts` field

fisker:
Remove support for `better-npm-run`, it's deprecated.

https://github.com/benoror/better-npm-run


keithamus:
"deprecated" doesnt mean no one uses it. I don't see it adding much maintainence burden to keep this in for a while. We can revisit in a few years perhaps?

Shall we close this?

fisker:
@keithamus sure, I'll close

PR #106, opened 2019-12-20
fix: add `contributors` field back

### LINKED ISSUE #105: People fields
[people fields](https://docs.npmjs.com/files/package.json#people-fields-author-contributors) include not only `author`, but `contributors` too.

Currently, only author in list:
https://github.com/keithamus/sort-package-json/blob/2108c5ade72463a6786b4408ea1dccbcaa6541fa/index.js#L73

fisker:
fixes #105


fisker:
// @fanatid

PR #107, opened 2019-12-20
docs: update README.md

fisker:
- Move CLI docs before API docs.
- Update examples.
- Document `options.sortOrder`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@
 
 ## CLI
 
+### Run via npx (npm@5.2+ required)
+
+```bash
+npx sort-package-json
+```
+
 ### Install
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,125 +1,171 @@
...
+$ sort-package-json "**/package.json" --check
 # foo/package.json
 # bar/package.json
 
 # 2 of 5 matched files are not sorted.
 ```
 
-
-### Run (with npm@5.2+)
+### Run with npx (npm@5.2+ required)
```
keithamus: Could we move this section higher? I dont think installing this package is the core use-case. `npx sort-package-json` is the ideal way to use this.

fisker: Moved to top of CLI section

PR #108, opened 2019-12-20
feat: prepend custom order to default sort order

fisker:
Make custom sortOrder prepend to defaultSortOrder, instead of override it, so user only need send keys that must on top, the rest known fields still sort by defaultOrder.
Should be much easier to use.


fisker:
forgot custom order could be function, should support?

keithamus:
We should support custom sort order being a function too.

fisker:
@keithamus I can add type check here, but are you ok with prepending custom sort order array to default array?

keithamus:
I think #102 subsumed this. Do you want to rebase for the test case at least?

keithamus:
I like how small this PR ended up being

PR #112, opened 2019-12-28
fix: sort `bundledDependencies` as array

fisker:
`bundledDependencies`/`bundleDependencies` should be an array, not object.

https://docs.npmjs.com/files/package.json#bundleddependencies


fisker:
Note to myself: `os` and `cpu` also array, will send seperate PR to fix them after `sortArray` function been added

PR #114, opened 2019-12-31
fix: correct type of `man` `os` `cpu` field

fisker:
`man` `os` `cpu` are all `array`. 

for `os` and `cpu`, they support negative values.
So, in theory it could be `['darwin', '!darwin']`, order should matters, we don't want break it. 
It could also be `['darwin', '!darwin', 'darwin']`, so `uniq` may also break it.

for `man` field, I'm not familiar with `man` command, not sure we should do anything.

@keithamus thoughts?


keithamus:
👍

PR #116, opened 2020-01-03
feat: sort `eslintConfig`

fisker:
Sort `eslintConfig` object, the ordering of key in my personal opinion, all keys can found [here](https://github.com/eslint/eslint/blob/acc0e47572a9390292b4e313b4a4bf360d236358/conf/config-schema.js).

But question is this too much for us?

If we start doing this kind of sorting, we may need sort other libs.
If not, we might want disable sorting for `eslintConfig`.


keithamus:
Happy to do this sorting. It's just a few more strings in the codebase so 👍 for me

PR #118, opened 2020-01-04
fix: do not crash on invalid data

fisker:
Make `sortPackageJson` accept any data and invalid string, leave them as it is.


fisker:
Make sense too.

PR #123, opened 2020-01-06
refactor: use `pipe` and `overProperty`

fisker:
`overProperty` function is from #122
`pipe` function is from #121

please merge them first

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,8 @@ const detectNewline = require('detect-newline').graceful
 const globby = require('globby')
 const gitHooks = require('git-hooks-list')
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -170,6 +175,11 @@ const fields = [
 ]
 
 const defaultSortOrder = fields.map(({ key }) => key)
+const overFields = pipe(
+  fields
+    .filter(({ over }) => over)
+    .map(({ key, over }) => overProperty(key, over)),
```
keithamus: a filter + map could be combined to a reduce.

fisker: fixed

PR #128, opened 2020-01-08
feat: sort `eslintConfig` deeper

fisker:
the testes become hard to write, should we have a test lib that supports snap testing?


keithamus:
We can give snapshot testing a go if you would like. I generally dislike it because it is usually treated as opaque and so developers just commit the new snapshots whenever they break, meaning they become useless. But I suppose we can have the discipline to check for this with each PR.

fisker:
`ava` `jest` or any other lib you prefer?

keithamus:
I'll leave this decision to you! Whatever you prefer!

PR #136, opened 2020-01-09
fix: only run `sortArray` and `uniq` on `string[]`

fisker:


package-lock.json
```diff
@@ -4584,7 +4584,7 @@
     },
     "is-obj": {
       "version": "1.0.1",
-      "resolved": "http://registry.npmjs.org/is-obj/-/is-obj-1.0.1.tgz",
```
fisker: How this http sneak in

PR #144, opened 2020-01-11
chore: run format on commit

fisker:
Setup `lint-staged` to format codes

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,13 +34,27 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm t",
+      "pre-commit": "npm t && lint-staged",
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,6 +43,20 @@
...
+    "*.js": [
+      "eslint --fix",
+      "git add"
+    ],
+    "*.ts": [
+      "prettier --write",
+      "git add"
+    ],
+    "package.json": [
+      "node cli.js",
```
keithamus: This is hilarious 😆

fisker: I tried to use stable `sort-package-json`, but seems npm can't install package with the same name? 

Should we swith to yarn?

keithamus: This is fine, it's like a little integration test for us 😄

fisker: Okay

PR #145, opened 2020-01-14
refactor: tiny code refactor

fisker:
1. make editStringJSON only accept string
2. move json parse to top, prepare for error handling

### DIFF
--- a/index.js
+++ b/index.js
@@ -259,23 +259,21 @@ const overFields = pipe(
   }, []),
 )
 
-function editStringJSON(json, over) {
-  if (typeof json === 'string') {
-    const { indent } = detectIndent(json)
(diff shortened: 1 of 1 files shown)


fisker:
I have a problem handling `non-package.json` file,

For example:

a json file, has content

```json
[
"a",
                        "b"
]
```

CLI read this file, and pass to `editStringJSON`, `JSON.parse` won't throw, because it's valid JSON string.

so pass to `over`, over will return it back, since it's not an object, but I can't throw here, because it will be a breaking change. and `sortPackageJson` did accept `array` without throwing any error, `array string` shouldn't throw either. 

But, `editStringJSON` return a new string, whitespace is changed. and CI has no way to know what happened.

fisker:
abandoned

PR #147, opened 2020-01-18
feat: add export default

bluelovers:
make sure import default can work


fisker:
Can you explain why we need this?

bluelovers:
```ts
import sortPackageJson from 'sort-package-json'
```

for use typescript when `esModuleInterop` not `true`

fisker:
I  thought people use 

```ts
import {sortPackageJson} from 'sort-package-json'
```

or

```ts
import sortPackageJson = require('sort-package-json')
```

bluelovers:
that 2 syntax not a good style in this case

PR #150, opened 2020-01-27
file should stay near dir

bluelovers:


bluelovers:
hmm i don't know how to update snapshot

fisker:
@bluelovers `npm run test -u`

package.json
```diff
@@ -17,20 +17,21 @@
...
-  "types": "index.d.ts",
-  "bin": "cli.js",
   "scripts": {
     "lint": "eslint .",
     "semantic-release": "semantic-release",
     "sort-package-json": "node cli.js package.json --check",
     "test": "ava",
-    "test-coverage": "nyc ava"
+    "test-coverage": "nyc ava",
+    "update-snapshot": "ava -u && node cli.js package.json"
```
fisker: I dont think we need this, its on `ava` docs

bluelovers: but this also help when update sort order, can more easy update package.json and snapshot at same time

keithamus: I like this. I’d probably forget `-u` and this is a nice reminder.

PR #153, opened 2020-01-27
chore: update `ava` to v3

fisker:
fixes: #148


package.json
```diff
@@ -70,9 +70,11 @@
     ]
   },
   "ava": {
+    "babel": true,
     "verbose": true
```
keithamus: We don’t need Babel right? Can this be disabled?

fisker: Done.

PR #158, opened 2020-02-18
feat: sort eslint rules

fisker:
Sort `eslintConfig.rules` in a better way.


keithamus:
I like the idea of sorting eslint rules, especially by project, but I think they should not be alphabetically sorted because often rules are grouped together. To give a concrete example; one might want to group array rules like `no-sparse-arrays` next to `no-array-constructor`.

fisker:
We  already sorting them alphabetically, so I should remove that?

keithamus:
Ah! Did not realise we were doing that. I see that was done in https://github.com/keithamus/sort-package-json/pull/116. I guess my reasoning isn't that consistent 😆

G-Rath:
I know this is close to bikeshedding, but is there a particular reason why plugins are sorted after recommended?

I prefer sorted like this rather than strictly alphabetically, but prefer the plugins first since they're longer and what tends to be more unique to specific projects (i.e all projects have access to the eslint rules, and so those are the rules that tend to be the most consistent).

fisker:
@G-Rath I think there is no `recommended` property in `eslintConfig`

G-Rath:
Sorry apologies I meant core rules.

fisker:
I didn't think too much, I thought  that's people usually do(and I do in this way). 

And try to check popular project code base, but not too much to see, but 

https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json this file does this way, I'll try to find some popular project not doing in this way

G-Rath:
Huh interesting - I thought it was the other way around, but can't name any repos off hand.

I'd be interested in seeing roughly what the ratio is, if any, if you don't mind exploring some metrics together 🙂

PR #169, opened 2020-04-04
feat: add remarkConfig field

jeddy3:
Adds [remark]() amongst ESLint and stylelint.

It is a popular package with [8M/monthly downloads](https://www.npmjs.com/package/remark). We use it to [lint the Markdown files](https://github.com/stylelint/stylelint/blob/0a8efea5e6e3bf8f1fc61492296669d1160ec4fd/package.json#L251-L261) across the stylelint repositories.

It [uses the `remarkConfig` property](https://github.com/remarkjs/remark-lint/blob/master/readme.md#cli).


fisker:
Should also tested in https://github.com/keithamus/sort-package-json/blob/master/tests/fields.js

jeddy3:
Thanks for the heads-up. I've added it to the `sortObjectAlphabetically` test in `fields.js`.

jeddy3:
It looks like the CI didn't trigger. Perhaps because the pull request comes from a fork?

keithamus:
Looks like travis-ci is dead :( we should move to GitHub Actions

fisker:
travis-ci is dead?

keithamus:
It looks like it's only running half the time

Issue #172, opened 2020-04-08
Default export has no call signature in types?


nemoDreamer:
After a long-overdue update from `^1.15.0` to `1.41.0`, I'm getting this when checking my Typescript types with `tsc`:

```
packages/scripts/normalize-packages.js:290:29 - error TS2349: This expression is not callable.
  Type 'typeof import("/Users/philip/Development/Autodesk/shotgun-ux/sgds/node_modules/sort-package-json/index")' has no call signatures.

290           `${JSON.stringify(sortPackageJSON(packageJSON), null, 2)}\n`,
                                ~~~~~~~~~~~~~~~

Found 1 error.
```

I'm importing via `const sortPackageJSON = require('sort-package-json');` and my script itself works fine.

keithamus:
This might be related to https://github.com/keithamus/sort-package-json/pull/170 /cc @bluelovers

bluelovers:
![image](https://user-images.githubusercontent.com/167966/78917348-56b1ef80-7ac1-11ea-8530-a58a099e9bfa.png)

didn't see error

nemoDreamer:
Sweet, thanks so much! ❤️

nemoDreamer:
@bluelovers after updating to `1.42.1`, I no longer get the warning in my IDE (VS Code), but get this when running `tsc` Typescript-check:

```
node_modules/sort-package-json/index.d.ts:14:7 - error TS1134: Variable declaration expected.
14   var default: <T extends Record<any, any>>(packageJson: T, options?: sortPackageJsonCore.Options) => T;
         ~~~~~~~
node_modules/sort-package-json/index.d.ts:14:14 - error TS1134: Variable declaration expected.
14   var default: <T extends Record<any, any>>(packageJson: T, options?: sortPackageJsonCore.Options) => T;
                ~
node_modules/sort-package-json/index.d.ts:14:16 - error TS1134: Variable declaration expected.
14   var default: <T extends Record<any, any>>(packageJson: T, options?: sortPackageJsonCore.Options) => T;
                  ~

Found 3 errors.
```

PR #173, opened 2020-04-09
fix: try avoid some typescript setting show error

bluelovers:
https://github.com/keithamus/sort-package-json/issues/172


Jamesgt:
It is not yet fixed.

node_modules/sort-package-json/index.d.ts:14:14 - error TS1134: Variable declaration expected.

14   var default: <T extends Record<any, any>>(packageJson: T, options?: sortPackageJsonCore.Options) => T;

Jamesgt:
Now it is fixed in #189 and version 1.42.3

PR #176, opened 2020-04-16
ci: move to github actions

fisker:


fisker:
You are a github staff at github, right? A off topic question, why actions only run after "master branch has a workflow"?

keithamus:
Right, I believe workflows only run if the master branch has a workflow of that name, due to security reasons.

fisker:
Wasted couple hours on initial setup in other repos.

Good to know, thanks.

PR #177, opened 2020-04-16
make release run as main job

keithamus:


.github/workflows/ci.yml
```diff
@@ -40,28 +40,10 @@ jobs:
...
-        - name: Release
-          run: npm run semantic-release
-          env:
-            GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-            NPM_TOKEN: ${{secrets.NPM_TOKEN}}
-
+        
+      - name: Release
+        if: github.ref == 'master'
+        run: npm run semantic-release
```
friederbluemle, line 46: @keithamus 
Just wondering.. This step is now part of the "test" job.. Will it be executed for each individual run in the build matrix?
Right now there are 4 runs for different Node.js versions kicked off for each build. When you add the other two operating systems, it'll be a total of 12 jobs!

friederbluemle: Hmm I guess right now this step is skipped. Is it because `if: github.ref == 'master'` evaluates to false?

keithamus: This needs a little more work to do the right thing, but yeah it's skipped based on the `if`.

PR #179, opened 2020-04-16
Update prettier to 2.0.4

friederbluemle:
This updates `prettier` to `2.0.4`. Note that the new version comes with several minor style updates as well, which requires code changes in the same commit (which is why #164 failed).

This also removes Node.js 8 from the GitHub Actions test matrix. @keithamus - If you require Node.js 8 compatibility, please close this PR (as well as #164). Node.js 8 is [no longer supported](https://nodejs.org/en/about/releases/), and Prettier 2 seems to be incompatible with it

Supersedes and closes #164

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
         run: npm ci
 
       - name: Run ESLint
+        if: matrix.node_version == '12'
         run: npm run lint
 
       - name: Check package.json
@@ -40,7 +41,7 @@ jobs:
 
       - name: Run Tests
         run: npm run test-coverage
-        
+
       - name: Release
         if: github.ref == 'master'
         run: npm run semantic-release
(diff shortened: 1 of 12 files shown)


fisker:
@friederbluemle You don't have to remove Node.js@8, just skip eslint by `if`.

friederbluemle:
@fisker That's even better. Thanks for the suggestion! :+1:

fisker:
Haha, FYI: you can still set  [`arrowParens` to `avoid`](https://prettier.io/docs/en/options.html#arrow-function-parentheses)

friederbluemle:
They are giving some additional details on this decision in their docs:

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.

Personally, in this specific case I don't have a strong preference for or against the parens. However, most of the time it's easier to go _with_ the defaults instead of _against_ them (convention over configuration).

PR #161, opened 2020-02-29
chore: update dependencies

friederbluemle:
Includes, supersedes, and closes #152 
Includes, supersedes, and closes #155  
Includes, supersedes, and closes #156
Includes, supersedes, and closes #157 
Includes, supersedes, and closes #160

PR #187, opened 2020-05-03
fix: undo semver breaking globby update

kellyselden:
I couldn't figure out what Node versions you support or if you use semver, but this came up in a couple projects. Globby 11 drops node 8, and this project was not major version bumped.

https://github.com/keithamus/sort-package-json/pull/161/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R77

https://github.com/sindresorhus/globby/releases/tag/v11.0.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -773,6 +773,22 @@
         "p-filter": "^2.0.0",
         "p-retry": "^4.0.0",
         "url-join": "^4.0.0"
+      },
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


fisker:
Isn't it "Require Node.js 10", you mean Node.js 8?

kellyselden:
Good catch. updated to "Globby 11 drops node 8"

PR #188, opened 2020-05-21
feat: update lifecycle logic to handle - after prefix

Larry1123:
I don't think that this should affect any. I am working on a plugin for yarn 2 to add lifecycle handing.
As they say in their docs using just `pre` and `post` can have odd side effects. The plugin I am creating will use `pre-` and `post-` as the prefix. I use this package and wanted to keep using it without patching it per project to make use of it.

From yarn docs.
`
In particular, we intentionally don't support arbitrary pre and post hooks for user-defined scripts (such as prestart). This behavior, inherited from npm, caused scripts to be implicit rather than explicit, obfuscating the execution flow. It also led to surprising executions with yarn serve also running yarn preserve.
`

### DIFF
--- a/index.js
+++ b/index.js
@@ -137,7 +137,7 @@ const sortScripts = onObject((scripts) => {
 
   const keys = names
     .map((name) => {
-      const omitted = name.replace(/^(?:pre|post)/, '')
+      const omitted = name.replace(/^(?:pre-|post-|pre|post)/, '')
       if (defaultNpmScripts.has(omitted) || names.includes(omitted)) {
         prefixable.add(omitted)
         return omitted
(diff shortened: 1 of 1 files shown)


Larry1123:
I am unsure why eslint is failing it is not failing locally when I run it.

fisker:
Any docs suggest `pre-` instead of `pre` ?

Larry1123:
No I have not. I had looked over a lot of names that come up and I felt that even if it was used most uses causes like `pre-commit` would likely be fine with being sorted near and around other commands like it anyway.
I was looking for a way to add a new way to do lifecycle hooks that would not match other known ones, but a naming format that would still work even if someone was not needing it.
The main reason I went with using `-` was that it would avoid issues with words inadvertently matching.

PR #199, opened 2020-09-08
feat: add support for Tap

revelt:
Tap is a synchronous unit test runner, written by Isaac Schlueter, creator of npm, and at the moment, it gets around 88K weekly downloads: https://www.npmjs.com/package/tap. For comparison, that's around 4 times the `xo` and half the `ava` download traffic and we support both here. Tap deserves to be mentioned among all other unit test runners.

Tap documentation says:

> "... tap will look for a tap section in a package.json file"
— https://node-tap.org/docs/configuring/


keithamus (approved):
:+1:

PR #203, opened 2020-10-19
Removed `exports` sort, added `imports`

### LINKED ISSUE #197: Leave conditional exports alone
https://nodejs.org/api/esm.html#esm_conditional_exports states

> Within the "exports" object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries. *The general rule is that conditions should be from most specific to least specific in object order.*

I found that `default` was sorted before `node`.

Xunnamius:
Closes #197


keithamus:
Looks good, but snapshot tests are failing. If you run `npm t` locally you can see them fail. If you need any assistance fixing them up @fisker can hopefully help!

fisker:
`npm run update-snapshot` should be able to fix it.

Xunnamius:
Thanks, I'll take a look at this tonight!

Xunnamius:
Fixed. I made the initial commit through the GitHub UI so I missed the commitlint pass. Let me know if I should amend the commit message

keithamus (approved):

PR #206, opened 2020-11-12
fix: stop sorting scripts

jonathanmorley:
The fields option will override the values in the builtin fields array,
allowing users to signify that a given key should not be sorted, or sorted
in a particular way.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -35,7 +35,6 @@ declare namespace sortPackageJson {
 
   interface Options {
     readonly sortOrder?: readonly string[] | ComparatorFunction
-    readonly fields?: Field[]
   }
 }
 
@@ -43,7 +42,6 @@ interface sortPackageJsonExports extends sortPackageJson.SortPackageJsonFn {
   readonly default: sortPackageJson.SortPackageJsonFn
   readonly sortPackageJson: sortPackageJson.SortPackageJsonFn
   readonly sortOrder: string[]
-  readonly sortObjectBy: typeof sortPackageJson.sortObjectBy
 }
 
 declare const sortPackageJsonExports: sortPackageJsonExports
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,36 +3,49 @@
 declare namespace sortPackageJsonExports {
   interface SortPackageJsonFn {
     /**
-      * Sort packageJson object.
-      *
-      * @param packageJson - A packageJson
-      * @param options
-      * @returns Sorted packageJson object
-      */
-    <T extends Record<any, any>>(packageJson: T, options?: Options): T,
+     * Sort packageJson object.
+     *
+     * @param packageJson - A packageJson
+     * @param options
+     * @returns Sorted packageJson object
+     */
+    <T extends Record<any, any>>(packageJson: T, options?: Options): T
 
     /**
-      * Sort packageJson string.
-      *
-      * @param packageJson - A packageJson string.
-      * @param options
-      * @returns Sorted packageJson string.
-      */
-    (packageJson: string, options?: Options): string,
+     * Sort packageJson string.
+     *
+     * @param packageJson - A packageJson string.
+     * @param options
+     * @returns Sorted packageJson string.
+     */
+    (packageJson: string, options?: Options): string
   }
 
-  type ComparatorFunction = (left: string, right: string) => number;
+  type ComparatorFunction = (left: string, right: string) => number
+  function sortObjectBy<T extends Record<any, any>>(
+    comparator: string[],
+    deep: boolean,
+  ): (x: T) => T
+
+  interface Field {
+    readonly key: string
+    readonly over: sortObjectBy
+  }
 
   interface Options {
-    readonly sortOrder?: readonly string[] | ComparatorFunction;
+    readonly sortOrder?: readonly string[] | ComparatorFunction
+    readonly fields?: Field[]
   }
 }
 
-interface sortPackageJsonExports extends sortPackageJsonExports.SortPackageJsonFn {
-  readonly default: sortPackageJsonExports.SortPackageJsonFn;
-  readonly sortPackageJson: sortPackageJsonExports.SortPackageJsonFn;
+interface sortPackageJsonExports
+  extends sortPackageJsonExports.SortPackageJsonFn {
+  readonly default: sortPackageJsonExports.SortPackageJsonFn
+  readonly sortPackageJson: sortPackageJsonExports.SortPackageJsonFn
+  readonly sortOrder: string[]
+  readonly sortObjectBy: sortPackageJsonExports.sortObjectBy
 }
 
-declare const sortPackageJsonExports: sortPackageJsonExports;
+declare const sortPackageJsonExports: sortPackageJsonExports
 
-export = sortPackageJsonExports;
+export = sortPackageJsonExports
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,12 +1,12 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-declare namespace sortPackageJsonExports {
+declare namespace sortPackageJson {
   interface SortPackageJsonFn {
     /**
      * Sort packageJson object.
      *
      * @param packageJson - A packageJson
-     * @param options
+     * @param options - An options object
      * @returns Sorted packageJson object
      */
     <T extends Record<any, any>>(packageJson: T, options?: Options): T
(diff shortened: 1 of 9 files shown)


index.d.ts
```diff
@@ -3,36 +3,49 @@
...
+     * @param packageJson - A packageJson string.
+     * @param options
+     * @returns Sorted packageJson string.
+     */
+    (packageJson: string, options?: Options): string
   }
 
-  type ComparatorFunction = (left: string, right: string) => number;
+  type ComparatorFunction = (left: string, right: string) => number
+  function sortObjectBy<T extends Record<any, any>>(
```
jonathanmorley, line 25: I wasn't quite sure how to test this, I am not used to hand-writing type definitions

fisker:
Why not extend `sortOrder`  to `['name', { key: 'repository', over: myOwnSortFunction }]`

jonathanmorley:
I started going down that route, but if you just specify something like `sortOrder: [{ key: 'scripts', over: myOwnSortFunction }]` then I wouldn't want that to cause the `scripts` object to appear at the top of the object.

This way you can influence sort of keys without affecting the ordering of those keys

fisker:
How about  `defaultSortOrder.map(key => key === "script" ? { key: 'scripts', over: myOwnSortFunction } : key)`

jonathanmorley:
While that would work, it seems more confusing to me. What is the reasoning behind preferring re-using the `sortOrder` option rather than introducing a new option?

index.d.ts
jonathanmorley: added tests

jonathanmorley:
what would be needed to move this forward?

jonathanmorley:
@keithamus ?

keithamus:
Could you please demonstrate a concrete use case for needing this?

keithamus:
@jonathanmorley I would prefer to keep `sort-package-json` configuration free. So if that is the only use case you're solving, we can:

1. Simply not sort `scripts`. This is my preferred option.
2. Detect `npm-run-all` in devdeps and conditionally not sort. I'm willing to entertain this idea with more exploration.

fisker:
> Detect npm-run-all in devdeps and conditionally not sort.

👍

jonathanmorley:
npm-run-all has a third command (`npm-run-all`) which also has the ability to run commands sequentially.

Rather than add in checks for `npm-run-all` or `run-s`, I opted for the approach 2 in https://github.com/keithamus/sort-package-json/pull/206#issuecomment-763473034

fisker:
What if `npm-run-all` is installed global.

nelson6e65:
Why not just an option to (dis)allow ordering scripts, regardless of the reason, instead of check for a specific tool? What if I use another tool like npm-run-all but is called different? I need to create another patch for each specific tool that requires specific order on scripts?

jonathanmorley:
that was the original intent. The `fields` option would allow a consumer to choose to disabling sorting, or choose a custom sort method for a provided field(s)

keithamus:
Yeah we can just unilaterally disable sorting of scripts. In my experience these tend to be more taken care of than the rest of the package.json, and the important part of this package is really to sort the main fields so you can predictably `head` the file to see stuff like `name` and `main`

matzkoh:
I use npm-run-all and prefer to sort the scripts.
For example, it makes sense for prefoo foo postfoo to be sorted in that order.

It would be a shame to stop sorting scripts just because some tools depend on the key order of the objects.
(Rather than "the tools depend on it", the users depend on the behavior that happens to work well?)

It is an object, not an array.

keithamus:
We could _just_ sort `pre` and `post` scripts, as we know them to have consistent behaviour with and without `npm-run-all`.

nelson6e65:
> 
> 
> We could _just_ sort `pre` and `post` scripts, as we know them to have consistent behaviour with and without `npm-run-all`.

I agree with having `pre`/`post` scripts sorted it is useful.

keithamus:
I'm finding this diff really difficult to read now. Why does it change a bunch of seemingly unrelated lines? Do you have auto formatting on?

index.js
```diff
@@ -135,16 +135,14 @@ const sortScripts = onObject((scripts) => {
...
-  const keys = names
-    .map((name) => {
-      const omitted = name.replace(/^(?:pre|post)/, '')
-      if (defaultNpmScripts.has(omitted) || names.includes(omitted)) {
-        prefixable.add(omitted)
-        return omitted
-      }
-      return name
-    })
-    .sort()
```
keithamus, line 147: Seems like this is the only LOC that should be in this diff now, no?

jonathanmorley: I believe this is prettier moving the `.map` up a line since it no longer has a `.sort` to align with

jonathanmorley: @keithamus does this sufficiently explain the diff?

keithamus: Why has the `index.d.ts` changed so much? Is it just that the copy in the main branch is old?

keithamus (approved):

G-Rath:
I'm a bit sad that this was merged, especially as a bug: having `scripts` sorted consistently across projects as a default is one of the things I love(d) about `sort-package-json`.

Would it be possible to revisit this to see if we could come up with a middleground for everyone?

joemaller:
I agree with @G-Rath, this change was an unwelcome surprise. We've been using sort-package-json to standardize our package.json files for several years. We use a lot of scripts and keeping those in order was a huge help. After a recent round of updates, I thought something broke when the scripts weren't sorted back into the expected lexical+pre/post order. 

I see sort-package-json as a linting tool, and much as I've given formatting over to tools like Prettier, I'm fine with whatever opinionated choices the tool's maintainers decide to make. But removing script sorting a major loss of functionality which I'll be looking to replace.

benquarmby:
> I agree with @G-Rath, this change was an unwelcome surprise.

I'd go so far as to call it a breaking change. We will use `~1.48.0` as an escape hatch for now.

DUBANGARCIA:
> I agree with @G-Rath, this change was an unwelcome surprise. We've been using sort-package-json to standardize our package.json files for several years. We use a lot of scripts and keeping those in order was a huge help. After a recent round of updates, I thought something broke when the scripts weren't sorted back into the expected lexical+pre/post order.
> 
> I see _sort-package-json_ as a linting tool, and much as I've given formatting over to tools like _Prettier_, I'm fine with whatever opinionated choices the tool's maintainers decide to make. But removing script sorting is a major loss of functionality which I'll be looking to replace. For now, running `npx sort-package-json@1.48` was a sigh of relief.

Totally agree, the ordering of scripts was very important for my projects to avoid unnecessary merges with git and improve the readability of the package.json, please reconsider this change.

Maxim-Mazurok:
I just want to let you know that it also was a surprise that our scripts are not being sorted and I just noticed this... Guess it is me to blame because I didn't pin this package version and was just using the latest.

But yeah, I agree with the guys above that this was a pretty important feature for me, prettier/linter for pacakge.json, in a way. And I'm sad to see this functionality gone. Will pin to `1.48` for now.

keithamus:
> This edge case really wasn't very hard to solve and costs very little

Please do submit a PR if you think this is solvable in a deterministic and non-configurable way that addresses the concerns of folks like @jonathanmorley.

benquarmby:
@keithamus : The deterministic way was covered in the code example. To be clear, here it is again:
```diff
- "build": "run-s build:*",
+ "build": "run-s build:ts build:cdk",
```
Being explicit about the order and avoiding the magical `*` is as deterministic as it gets. I say this as someone who makes heavy use of `npm-run-all`.

I'll be blunt: `sort-package-json` should never have been watered down to cater for dynamic usage of `npm-run-all`. It was a breaking change for my projects.

keithamus:
@G-Rath has submitted a PR which sorts scripts _unless_ `npm-run-all` is installed.

nelson6e65:
Hmm... What if I have installed and still want to sort them? Examples above shows that.

Maxim-Mazurok:
> Hmm... What if I have installed and still want to sort them? Examples above shows that.

I guess you have two choices:
1. Install it as a regular (non-dev) dependency. Should be fine for non-library projects.
2. Uninstall it and use through `npx`

keithamus:
If you would like to improve the heuristic - if you think you could find a deterministic way to sort scripts without effecting existing implicit `npm-run-all` script orders, then PRs are welcome! For example we could check the contents of scripts to determine if they're using `run-s` with a wildcard.

fisker:
Don't forget `npm-run-all foo:* -s`,  `npm-run-all --sequential foo*` and other cases.

kachkaev:
Not sure if this is relevant, but `npm-run-all` has been unmaintained since October 2019. I tried bumping dependencies to get rid of known vulnerabilities, but [my PR](https://github.com/mysticatea/npm-run-all/pull/205) did not get any attention so far.

We might need to implement a rule that is flexible, i.e. not hard-wired to `npm-run-all`.

saiichihashimoto:
Went through a "bisect" of sort-package-json versions until I ended up here to determine this was not a bug, but a conscious decision. This is definitely a breaking change and magic sorting (or lack of sorting?) because a dependency is included is unpredictable behavior. I use `npm-run-all` extensively and the lack of sorting is painful.

Users of sort-package-json should assume that the json will be resorted (excluding arrays); relying on ordering of object keys is not a bug within sort-package-json, but relying on an unreliability. `sort-package-json` or `npm-run-all` should not consider this an issue, considering `npm-run-all` gives escape hatches for this. If there's a manual order you need scripts to be run, then manually run them in that order. `run-s *` and `run-p *` should be agnostic to the order they're run.

PR #211, opened 2020-11-26
feat: keep comment-hack keys above preceding key

### LINKED ISSUE #1: feature request comment-hack at the top?
First: I just stumbled on this and fully satisfies my neatly needs, thanks!

But I used a hack from:
http://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install#14221781
to make a little comment in my package.json (some dependency does not yet work in Windows, but they are working on it)
It would be nice to see the "//": "comment" above all others when it is used?

forivall:
Fixes #1 

the appropriate formatting of comments should be to put most of your comment in the key itself, like:

```
  "// Check with maintainer before adding anything!!": null,
  "dependencies": {
    ...
```

This is still a work in progress to add the functionality for nested things (like the `scripts` field), but it passes the test I added, so :tada:


index.js
```diff
@@ -317,21 +320,53 @@ const partition = (array, predicate) =>
...
+          : result[2]
+        : result[0]
+      ).unshift(key)
+      return result
+    },
+    // 0: non-comment keys and trailing comments
+    // 1: comments that precede a key
+    // 2: trailing comments
+    [[], {}, []],
+  )
```
fisker, line 339: This seems too complicated to me.

```js
['//z', '//x', 'z', ' //name', 'name', 'a', '// end']
 -> keys without `//`: ['z', 'x', 'name', 'a', 'end']
 -> sort: ['a', 'end', 'name', 'x', 'z']
 -> restore: ['a', `// end`, '//name', name, '//x',''x', '//z', 'z' ]
```

WDYT?

forivall: if anyone can think of simpler ways to do this, i'm all ears.

PR #230, opened 2021-08-30
feat: support `c8`

fisker:
https://github.com/bcoe/c8#cli-options--configuration

> c8 can be configured via command-line flags, a c8 section in package.json, or a JSON configuration file on disk.

I put it together with `nyc`, so  we don't need to move it when switch `nyc` to `c8`


fisker:
@keithamus Friendly ping.

I'm switching `nyc` to `c8`, I'd like to see this update to avoid unnecessary diff.

keithamus (approved):
LGTM :+1:

fisker:
Thanks!

PR #232, opened 2021-09-22
feat: sort all scripts unless `npm-run-all` is a dependency

### LINKED ISSUE #220: Script not sorting in 1.49.0
Pacakge.json scripts prop does not get sorted at all.

e.g remains the same order

```
{
   "scripts": {
        "b": "b",
        "a": "a"
   }
}
```

G-Rath:
This restores the default behaviour of sorting scripts alphabetically followed by pre & post _unless_ `npm-run-all` is a development dependency in the `package.json` being sorted.

Closes #220

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ const isPlainObject = require('is-plain-obj')
 const hasOwnProperty = (object, property) =>
   Object.prototype.hasOwnProperty.call(object, property)
 const pipe = (fns) => (x, ...args) =>
-  fns.reduce((result, fn) => fn(x, ...args), x)
+  fns.reduce((result, fn) => fn(result, ...args), x)
 const onArray = (fn) => (x) => (Array.isArray(x) ? fn(x) : x)
 const onStringArray = (fn) => (x) =>
   Array.isArray(x) && x.every((item) => typeof item === 'string') ? fn(x) : x
--- a/index.js
+++ b/index.js
@@ -6,14 +6,16 @@ const isPlainObject = require('is-plain-obj')
 
 const hasOwnProperty = (object, property) =>
   Object.prototype.hasOwnProperty.call(object, property)
-const pipe = (fns) => (x) => fns.reduce((result, fn) => fn(result), x)
+const pipe = (fns) => (x, packageJson) =>
+  fns.reduce((result, fn) => fn(result, packageJson), x)
 const onArray = (fn) => (x) => (Array.isArray(x) ? fn(x) : x)
 const onStringArray = (fn) => (x) =>
   Array.isArray(x) && x.every((item) => typeof item === 'string') ? fn(x) : x
 const uniq = onStringArray((xs) => xs.filter((x, i) => i === xs.indexOf(x)))
 const sortArray = onStringArray((array) => [...array].sort())
 const uniqAndSortArray = pipe([uniq, sortArray])
-const onObject = (fn) => (x) => (isPlainObject(x) ? fn(x) : x)
+const onObject = (fn) => (x, packageJson) =>
+  isPlainObject(x) ? fn(x, packageJson) : x
 const sortObjectBy = (comparator, deep) => {
   const over = onObject((object) => {
     object = sortObjectKeys(object, comparator)
@@ -38,9 +40,9 @@ const sortDirectories = sortObjectBy([
   'example',
   'test',
 ])
-const overProperty = (property, over) => (object) =>
+const overProperty = (property, over) => (object, packageJson) =>
   hasOwnProperty(object, property)
-    ? Object.assign(object, { [property]: over(object[property]) })
+    ? Object.assign(object, { [property]: over(object[property], packageJson) })
     : object
 const sortGitHooks = sortObjectBy(gitHooks)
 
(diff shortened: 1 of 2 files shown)


Maxim-Mazurok:
Doesn't cover `npx npm-run-all` case, but having `npm-run-all` as a dev dependency seems like a reasonable requirement in case if I don't want my scripts to be sorted. LGTM 👍

fisker:
The idea is good, but the implementation is bad, we should not pass `packageJson` all the way around.

I would check `npm-run-all` in `sortPackageJson` function and created a different `overFields` function.

G-Rath:
@fisker I'm not sure if I'd say that's a better solution - the advantage I see with passing `packageJson` around is that it can be used elsewhere for other advanced checks and it's implementation does that in a very clean way by encapsulating the special logic within the function that handles sorting scripts so other functions don't need to know about that.

If you were to make a different `overFields` function, if I understand you correctly, that'd result in something more complex and special-case-y since it would be devoted purely to the scripts fields, e.g. how would you handle a second independent field with similar behaviour? You'd have a third, maybe even four different `overFields` functions?

That's if I understand your thinking correctly - I've not done a lot of work with this sort of functional currying & piping layout so it's possible I've missed an easy trick :)

index.js
```diff
@@ -6,14 +6,16 @@ const isPlainObject = require('is-plain-obj')
 
 const hasOwnProperty = (object, property) =>
   Object.prototype.hasOwnProperty.call(object, property)
-const pipe = (fns) => (x) => fns.reduce((result, fn) => fn(result), x)
+const pipe = (fns) => (x, packageJson) =>
+  fns.reduce((result, fn) => fn(result, packageJson), x)
```
fisker, line 10:
```suggestion
const pipe = (fns) => (...args) =>
  fns.reduce((result, fn) => fn(...args), x)
```

index.js
```diff
@@ -6,14 +6,16 @@ const isPlainObject = require('is-plain-obj')
...
 const onArray = (fn) => (x) => (Array.isArray(x) ? fn(x) : x)
 const onStringArray = (fn) => (x) =>
   Array.isArray(x) && x.every((item) => typeof item === 'string') ? fn(x) : x
 const uniq = onStringArray((xs) => xs.filter((x, i) => i === xs.indexOf(x)))
 const sortArray = onStringArray((array) => [...array].sort())
 const uniqAndSortArray = pipe([uniq, sortArray])
-const onObject = (fn) => (x) => (isPlainObject(x) ? fn(x) : x)
+const onObject = (fn) => (x, packageJson) =>
+  isPlainObject(x) ? fn(x, packageJson) : x
```
fisker, line 18:
```suggestion
const onObject = (fn) => (...args) =>
  isPlainObject(x) ? fn(...args) : x
```

index.js
```diff
@@ -38,9 +40,9 @@ const sortDirectories = sortObjectBy([
   'example',
   'test',
 ])
-const overProperty = (property, over) => (object) =>
+const overProperty = (property, over) => (object, packageJson) =>
   hasOwnProperty(object, property)
-    ? Object.assign(object, { [property]: over(object[property]) })
+    ? Object.assign(object, { [property]: over(object[property], packageJson) })
     : object
```
fisker, line 46:
```suggestion
const overProperty = (property, over) => (object, ...args) =>
  hasOwnProperty(object, property)
    ? Object.assign(object, { [property]: over(object[property], ...args) })
    : object
```

fisker (commented):
> that'd result in something more complex and special-case-y since it would be devoted purely to the scripts fields

You are right, I was misleading by the parameter name, what do you think my suggestions?

G-Rath:
> what do you think my suggestions?

I'm not too opposed to them, but am interested in what others think as I originally did have the code like that, but the problem I saw was that to me it means you don't have this strong guarantee that the parameter is the package json that was passed in which is sort of the underlying idea: that a field is ensured it'll be passed whatever the value of the field is _and_ the whole package json object.

For me this is the sort of thing where I'd feel more comfortable using a general spread if I was writing TypeScript because I know that'll be able to ensure I'm getting passed the right properties for the field function I'm writing but when writing plain Javascript I prefer having that extra distention.

Additionally, this is all internal API - so shipping it how it is right now shouldn't later prevent switching to using a spread if the becomes a reason to sometimes pass more than one argument or an argument that isn't the whole package json object.

tl;dr: I don't know if there's a lot of value in using spread right now when we're always passing in the same two values, vs the extra clarity of using named parameter 🙂

Maxim-Mazurok:
I don't think that either approach is too bad. I don't think that dragging json is too bad for perf. But maybe better for ease of understanding. Either way, I haven't contributed to this package, so I'm going let you guys decide. I'm just excited to see #220 fixed asap :)

keithamus (commented):
I like the concepts in the code, I also agree with @fisker's assessment and code comments. Let's keep the "lib" stuff generic and then :ship:

G-Rath:
@keithamus are you ok to just commit the suggestions made by @fisker that you want made, since that should be all that's needed but I'm not at my laptop right now 😅

keithamus (approved):

bcomnes:
Posting here too for visibility: You can also add numerics to your script names to force a specific sort order while retaining the ability to automatically alphanumeric sort. IMO this is preferable than adding tool specific behaviors.

PR #240, opened 2022-01-21
feat: allow script sort for run-p

brodyd795:
### Background

This package does not currently support sorting the `scripts` field of a `package.json` when `npm-run-all` is a devDependency, regardless of how it is actually used in the scripts.

While there is certainly a place for refraining from sorting the scripts when `npm-run-all` is used for running scripts sequentially (e.g. `npm-run-all --serial`), there are use cases where the user may wish for the scripts to be sorted even when `npm-run-all` is a devDepenency (e.g. `npm-run-all -p`). This concern has been raised in previous PRs and issues (see [this PR](https://github.com/keithamus/sort-package-json/pull/206) and [this issue](https://github.com/keithamus/sort-package-json/issues/220)). 

**This PR enables this behavior.**

### Changes

The following uses of `npm-run-all` will continue to NOT sort the `scripts`:
- `npm-run-all`
- `npm-run-all --serial`
- `npm-run-all --sequential`
- `npm-run-all -s`
- `run-s`

While the following uses now WILL sort the `scripts`:
- `npm-run-all --parallel`
- `run-p`
-

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,6 +139,29 @@ const hasDevDependency = (dependency, packageJson) => {
   )
 }
 
+const runSValues = [new RegExp('run-s')]
+
+const hasRunS = (packageJson) => {
+  if (hasDevDependency('npm-run-all', packageJson)) {
+    const scripts = packageJson.scripts
+    const betterScripts = packageJson.betterScripts
+
+    if (scripts) {
+      return Object.values(scripts).some((script) =>
+        runSValues.some((runSValue) => runSValue.test(script)),
+      )
+    }
+
+    if (betterScripts) {
+      return Object.values(betterScripts).some((script) =>
+        runSValues.some((runSValue) => runSValue.test(script)),
+      )
+    }
+  }
+
+  return false
+}
+
 const sortScripts = onObject((scripts, packageJson) => {
   const names = Object.keys(scripts)
   const prefixable = new Set()
@@ -152,7 +175,7 @@ const sortScripts = onObject((scripts, packageJson) => {
     return name
   })
 
-  if (!hasDevDependency('npm-run-all', packageJson)) {
+  if (!hasRunS(packageJson)) {
     keys.sort()
   }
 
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
@keithamus Is there anything I can do to help in order to get this one merged?

keithamus:
Sure! I hadn't noticed this PR before as it's a draft and so didn't end up in my inbox. So to start I'd say taking this PR and making it a non-draft PR would be a good first step!

MichaelDeBoey:
> taking this PR and making it a non-draft PR would be a good first step!

@brodyd795 Can we set this PR 'ready for review' please?

index.js
```diff
@@ -139,6 +139,29 @@ const hasDevDependency = (dependency, packageJson) => {
...
+    }
+
+    if (betterScripts) {
+      return Object.values(betterScripts).some((script) =>
+        runSValues.some((runSValue) => runSValue.test(script)),
+      )
+    }
+  }
+
+  return false
```
MichaelDeBoey, line 162: What about an early return here?

```suggestion
  if (!hasDevDependency('npm-run-all', packageJson)) {
    return false
  }

  const scripts = packageJson.scripts
  const betterScripts = packageJson.betterScripts

  if (scripts) {
    return Object.values(scripts).some((script) =>
      runSValues.some((runSValue) => runSValue.test(script)),
    )
  }

  if (betterScripts) {
    return Object.values(betterScripts).some((script) =>
      runSValues.some((runSValue) => runSValue.test(script)),
    )
  }
```

index.js
```diff
@@ -163,8 +186,9 @@ const sortScripts = onObject((scripts, packageJson) => {
       ),
     [],
   )
+  const toReturn = sortObjectKeys(scripts, order)
 
-  return sortObjectKeys(scripts, order)
+  return toReturn
```
MichaelDeBoey, line 191: Since we're only using `toReturn` as a return value, I think we can just keep it as it was?

```suggestion
  return sortObjectKeys(scripts, order)
```

keithamus:
@MichaelDeBoey given this PR is coming on for a year old, you might consider forking it and making a new PR. You can keep @brodyd795's commits in, which means you'll both get contribution credit!

Issue #242, opened 2022-01-27
Enable sorting for npm-run-all


saiichihashimoto:
I've seen the PRs and reasoning for bailing on script sorting when npm-run-all is involved, but having the behavior drastically change based on the existence of a package is reasonably unexpected and should have been a breaking change, at the least. It makes a lot more sense to have a cli flag for `--disable-script-sorting` rather than making the behavior impossible if I'm using `npm-run-all`, so our team is stuck at version 1.48.1 indefinitely. Anyone who's using `sort-package-json` won't expect this sort of behavior and it's a surprise that can go undetected for months after introducing `npm-run-all`.

keithamus:
See https://github.com/keithamus/sort-package-json/pull/240 for a potentially better heuristic.

Also see https://github.com/keithamus/sort-package-json/issues/233#issuecomment-927639091 for why I do not wish this project to have any configuration flags.

saiichihashimoto:
https://github.com/keithamus/sort-package-json/issues/233#issuecomment-927639091 makes tons of sense to me; my qualm was that a semver major breaking change was made without notification or mitigation path, which was frustrating since it wasn't immediately obvious to us.

#240 is still unpredictable behavior. A specific package will disable script sorting, unless you only use it a specific way. From what I understand, https://github.com/keithamus/sort-package-json/issues/233#issuecomment-927639091 has the same methodology as prettier; you don't want to bikeshed over opinion where it truly doesn't matter and adding config just adds complexity. I'd argue that this `npm-run-all` behavior is encoding _opinion_ that **does** matter. This isn't the difference between tabs and spaces; this is the difference between enabling and disabling large chunks of functionality. We switch between serial and parallel for performance reasons, but now I have to couple that decision with whether I want scripts sorted or not. I can't justify the "either we run scripts in serial or have our scripts sorted" decision.

keithamus:
> same methodology as prettier

Yes sort of, except that prettier takes configuration now. Also in my opinion the lack of configuration (and thus determinism) of this project is fundamental to its benefit...

> you don't want to bikeshed over opinion where it truly doesn't matter and adding config just adds complexity

The intent of me making this tool is that I can open a package json and always expect to see (for example) `name` at the top. If people can override that determinism, then this tool is useless to me. Bikeshedding and complexity are somewhat afterthoughts here. 

> I'd argue that this `npm-run-all` behavior is encoding _opinion_ that **does** matter

[...]

PR #244, opened 2022-02-03
feat: sort regardless of devDependencies

saiichihashimoto:
This package does not currently support sorting the scripts field of a package.json when `npm-run-all` is a devDependency.

This PR proposes that `sort-package-json` should sort package.json, not create these exceptions, and revert those changes. They are unpredictable and hard to detect even with proper documentation. `npm-run-all` is likely not the only cli that can be concerned with script order and handling more of them isn't feasible. This has been brought up many times (#206, #220, #240, #242) and, since there are ultimately two conflicting schools of thought with no mitigation that allows both to exist, requires `sort-package-json` to pick one way or the other.


saiichihashimoto:
As mentioned in [this comment](https://github.com/keithamus/sort-package-json/issues/242#issuecomment-1028930036), we won't be doing this. Closing.

PR #255, opened 2022-04-03
chore!: update `globby` to v11

### LINKED ISSUE #247: Please update `globby@10.0.0` to at least 11.0.0
### Brief
sort-package-json@1.54.0 problem located

using globby@10.0.0
https://github.com/keithamus/sort-package-json/blob/v1.54.0/package.json#L77

### History
globby@10.0.0 has problem "Including @types/glob in dependencies breaks typescript implementations"

https://github.com/sindresorhus/globby/issues/129

Fixed in "Get rid of glob dependency"
https://github.com/sindresorhus/globby/pull/135

> Drop glob and @types/glob dependencies

https://github.com/sindresorhus/globby/commits/v11.0.0

### Comparison: package.json
globby@10.0.0
https://github.com/sindresorhus/globby/blob/v10.0.2/package.json#L59

globby@11.0.0
https://github.com/sindresorhus/globby/blob/v11.0.0/package.json#L59

### Related
https://github.com/microsoft/TypeScript/issues/48341

fisker:
Fixes #247.

BREAKING: requires Node.js >=12


keithamus:
As this PR is breaking, let's take this PR as an opportunity to upgrade all modules that drop Node 8 support, and look into if we can upgrade to ES Modules as part of that.

PR #256, opened 2022-04-04
refactor!: migrate to esm and update dependencies

fisker:
Fixes #254
Fixes #247
Fixes #227
Closes #255

[BREAKING]: Require Node.js 12

PR #259, opened 2022-05-01
Revert "refactor!: migrate to esm and update dependencies (#256)"

MichaelDeBoey:
This reverts commit 3e4ff74ea4126a417c7e5aff6d6d9d8a4f7c8b8a.

#256 should have ben a breaking/major release.

CC/ @fisker @keithamus

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -26,10 +26,6 @@ jobs:
         with:
           node-version: ${{ matrix.node_version }}
 
-      # https://github.com/bahmutov/npm-install/issues/103#issuecomment-931226602
-      - name: Update NPM
-        run: npm install --global npm
-
       - name: Install Dependencies
         run: npm ci
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,9 +12,6 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup Node.js
         uses: actions/setup-node@v3
-      # https://github.com/bahmutov/npm-install/issues/103#issuecomment-931226602
-      - name: Update NPM
-        run: npm install --global npm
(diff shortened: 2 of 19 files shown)


keithamus (approved):
Quite right. This one slipped through. Thanks!

ph-fritsche:
I just happened to have this contribution in my Feed. Have a look at [ph-fritsche/action-release](https://github.com/ph-fritsche/action-release).
It's a GitHub Action wrapper for `semantic-release` and it [defaults to Conventional Commits](https://github.com/ph-fritsche/action-release#defaults), i.e. `refactor!:` triggers a major release.

MichaelDeBoey:
@ph-fritsche What's the difference with [`cycjimmy/semantic-release-action`](https://github.com/cycjimmy/semantic-release-action)?

ph-fritsche:
First of all it's written in Typescript, is unit-tested and dogfeeds itself. I initially wanted to write a PR for cycjimmy's code, but wasn't confident to implement new features without tests or types so I started from scratch.

* There is a [default config](https://github.com/ph-fritsche/action-release/blob/master/src/defaultConfig.ts) that sets the `preset` to `conventionalcommits`, adds `main` to the release branches and omits the `npm` and `github` plugin if you omit the respective token.

* It installs the appropriate `conventional-changelog` package and any plugins you list in an inline or shared config.

* It adds and updates moving tags, i.e. for a `v1.2.3` it also moves `v1.2` and - if there isn't already a `v1.3` - `v1` to that commit, or for a `v1.2.3-alpha.4` it moves the `v1.2.3-alpha` tag.

fisker:
Should we revert it back? The original PR is fine, only  because the release bot not work as expected, we should revert this one and make sure we release v2.

PR #260, opened 2022-05-02
feat: add svelte to sort order list

mrtnbroder:
This keeps svelte entry points on the same level as main, unpkg, module etc.


keithamus (approved):

keithamus:
This will need some tests, I believe.

mrtnbroder:
@keithamus done!

keithamus (approved):

PR #266, opened 2022-07-18
feat: add support for `prettierIgnore`

mataha:
The goal of this PR is to add support for `prettierIgnore` - an alternative to defining ignored paths in [Prettier](https://prettier.io/) in a separate `.prettierignore` file, consisting of an array of file blobs that should *not* be sorted. The key behaves exactly the same as `eslintIgnore` does (with the exception of, well... handling ignores for Prettier instead of for [ESLint](https://eslint.org/)).

The PR in Prettier hasn't been merged yet, though I imagine it will be in the nearest future.

See: prettier/prettier#12672


fisker:
As a maintainer of this repo and Prettier, I have to say that Prettier PR is far from ready.

mataha:
> As a maintainer of both this repo and Prettier, I have to say that Prettier PR is far from ready.

Probably not the best place to have this conversation, but: what is it lacking, then?

fisker:
I'll review in Prettier now.

PR #269, opened 2022-09-28
refactor: migrate to esm and update dependencies

fisker:
Reverts keithamus/sort-package-json#259

Revert #256 back.

Please make sure we have a commit message footer this time https://github.com/keithamus/sort-package-json/pull/256#issuecomment-1087712876


keithamus (approved):
LGTM, we perhaps want to release this as semver MAJOR to avoid too much disruption from folks using the current version with older versions of Node

fisker:
v2.0.0 🚀

PR #274, opened 2022-10-09
feat: support `tsd`

bluelovers:


keithamus (approved):
This will need some tests, but otherwise :+1:

PR #275, opened 2022-11-04
feat: support jest plugins 'jest-junit' and 'jest-stare'

jfairley:
Adds support to sort the configuration for jest plugins `"jest-junit"` and `"jest-stare"` next to the main `"jest"` config.

- https://github.com/jest-community/jest-junit
- https://github.com/dkelosky/jest-stare


jfairley:
I don't know if there are other jest plugins worth considering. I think ideally there'd be a wildcard rule like `jest-*` that lined up right after `jest`, but I didn't want to get too deep in. 

Minimally, this will satisfy my own project needs.

keithamus:
> Minimally, this will satisfy my own project's needs.

That’s exactly the contributions I get excited about! Let’s not worry about theoretical gains. Scratch your own itch and someone else can come in after with more.

keithamus (approved):

PR #277, opened 2022-11-22
feat: allow sort `scripts` without `run-s`

MichaelDeBoey:
Re-submission of #240

CC/ @keithamus @fisker


index.js
```diff
@@ -148,6 +148,28 @@ const hasDevDependency = (dependency, packageJson) => {
   )
 }
 
+const runSValues = [/run-s/]
```
fisker, line 151: `npm-run-all -s`
`npm-run-all --sequential`
`npm-run-all --serial`

should also detected

index.js
```diff
@@ -148,6 +148,33 @@ const hasDevDependency = (dependency, packageJson) => {
   )
 }
 
+const runSValues = [
+  /npm-run-all -s/,
```
fisker, line 152: `/npm-run-all .* -s/`

fisker: Or more strict `/npm-run-all .* -s(?:$|\s|&|>|<)/`?

MichaelDeBoey:
@fisker I don't think the Node 12 is about something I've done in this PR, so this one can be merged if everything seems fine to you?

fisker:
So, you really think my suggestion should only apply to just one?

keithamus:
The node failures we can look into. We may need to drop node 12 support shortly.

MichaelDeBoey:
> So, you really think my suggestion should only apply to just one of the patterns?

@fisker Oh I thought it was meant to only have it on one of the regexes.
Since I'm not that good with regexes, can you please create a change request for the others as well?

> The node failures we can look into. We may need to drop node 12 support shortly.

@keithamus The problems with Node v12 CI is because we're always updating to latest `npm` version, which is v9 & is not supporting Node 12 anymore.
If we remove the "always update to latest version" step, Node 12 is perfectly supported & CI will pass

keithamus:
If you could please rebase, I've dropped that step in CI.

MichaelDeBoey:
@keithamus CI is passing on Node 12 now 👍

keithamus (approved):
LGTM but I'll let @fisker get a final look before merge.

fisker:
I need more time, but I have to go outside now, will be back in few hours.

fisker (approved):

MichaelDeBoey (commented):
@fisker It's probably out of scope for this PR, but what would you think about adding support for sorting all scripts, except the one's mentioned by `run-s`/`-s`/`--sequential`?

So basically

```json
{
  "scripts": {
    "test": "jest",
    "foo": "lint",
    "build": "build:*",
    "build:foo": "npm run foo",
    "build:bar": "npm run bar",
    "bar": "npm run test"
  }
}
```

will become



```json
{
  "scripts": {
    "bar": "npm run test",
    "build": "build:*",
    "build:foo": "npm run foo",
    "build:bar": "npm run bar",
    "foo": "lint",
    "test": "jest"
  }
}
```

fisker:
> except the one's mentioned by run-s/-s/--sequential?

Good, but maybe hard to implement.
Personally I  never run sequential tasks with `*`, only parallel tasks.

MichaelDeBoey:
In my personal projects I don't do it either, but I've come across it in open source & work projects

For example: when using Remix, it's important that `build:css` is ran before `build:remix` in the Tailwind case: https://remix.run/docs/en/v1/guides/styling#tailwind-css

Don't want to hold back on this PR for that though, so for me this one is approved as well 👍 (can't do it on GitHub, as I created the PR)

fisker:
@keithamus Let's merge this?

keithamus (approved):

PR #281, opened 2023-01-21
feat(cli): add --quiet / -q option

### LINKED ISSUE #280: Quiet CLI option to oppress success output
Adding a CLI option for to silence the output on success would be nice.

I am more than willing to submit a PR, however I figured I should ask a style opinion first. 
- A simple way to do this would be to override the console.log function with an empty arrow function `x=>{}`. The default console.log returns nothing, so this would be indistinguishable for the code. Errors could still be reported with console.error.
- The other option is to create an if check before each console.log for the quiet flag
Either way, this would be a quick addition that I'm more than happy to add. Please just let me know which direction you would prefer @keithamus.

aarondill:
This PR creates a --quiet (-q) option that simply stops the CLI from outputting success messages. Node errors and file-not-found errors are still outputted.

Tests have not been written, because the current testing infrastructure doesn't seem to have a way to automatically test the stdout of a CLI call.
closes #280


keithamus:
I would prefer an if check, or alternatively making a log function which encapsulates both the if check and the console.log statement. Overriding globals can cause confusion or maintenance issues.

aarondill:
I have made this change in my updated PR at #283

aarondill:
This has been reopened as per @fisker's request. It has been updated to match the master branch + my modifications and should now be ready for review.

cli.js
```diff
@@ -7,9 +7,31 @@ const isCheckFlag = (argument) => argument === '--check' || argument === '-c'
...
+
+function stdout(outputIfTTY = '', alwaysOutput = outputIfTTY) {
+  if (isQuiet) return
+  const isTerminal = !!process.stdout.isTTY
+  if (isTerminal) {
+    console.log(outputIfTTY)
+  } else if (alwaysOutput !== null) {
+    console.log(alwaysOutput)
+  }
+}
```
fisker, line 24: I can't understand why are we checking `process.stdout.isTTY` and what's the purpose of the `alwaysOutput` parameter.

Isn't this enough?

```js
const stdout = isQuiet ? () = {} : console.log
```

aarondill: You could do that for absolute output, I was planning for further changes, but I'm now going to move those to a new PR. I will revert this and stderr to a simpler form.

cli.js
```diff
@@ -35,7 +42,9 @@ if (isVersion) {
   process.exit(0)
 }
 
-const patterns = cliArguments.filter((argument) => !isCheckFlag(argument))
+const patterns = cliArguments
+  .filter((argument) => !isCheckFlag(argument))
+  .filter((argument) => !isQuietFlag(argument))
```
fisker, line 47: We can filter this with one call.

fisker:
Can you add test for those flags?

aarondill:
I'm currently working on this on a separate branch. Once I finish, I can copy the relevant tests to this branch.

aarondill:
Tests have been added

fisker (approved):

aarondill:
@fisker any update on this PR? I saw that you approved it, are we waiting for @keithamus to approve it or is there another holdup that I don't know about?

fisker:
Let's wait for @keithamus to take a look.

keithamus (approved):
LGTM :+1: nice work @aarondill :tada: 

Thanks @fisker for reviewing and helping get this into great shape!

PR #282, opened 2023-01-21
feat(cli): added --version and --help options

aarondill:
added --version (-V) and --help (-h) options to the CLI for user interface.

## `--help` output:
```
Usage: sort-package-json [OPTION...] [FILE...]
Sort npm package.json files. Default: ./package.json
Strings passed as files are parsed as globs.

  -c, --check                 check if FILES are sorted
  -q, --quiet                  don't output success messages
  -h, --help                   display this help and exit
  -V, --version              display the version and exit
  
```
This help menu can be easily modified in cli.js. It includes the option from #281, but that could be easily removed if so desired.

## `--version` output:
```
sort-package-json 0.0.0-development
```
Or other such versions. The version is pulled from the project.json file in the cli script's directory.
This is found by reading process.env[1], using fs.realpathSync, to read symlinks when it's installed, and path.dirname to get the parent directory. From there it is simple to read the file and call JSON.parse.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -30,9 +30,13 @@ Strings passed as files are parsed as globs.
 
 if (isVersion) {
   try {
+    const path = await import('node:path')
+
     const cliRealPath = fs.realpathSync(process.argv[1])
-    const cliParentDir = (await import('node:path')).dirname(cliRealPath)
-    const packageJsonBuffer = fs.readFileSync(`${cliParentDir}/package.json`)
+    const cliParentDir = path.dirname(cliRealPath)
+    const packageJsonBuffer = fs.readFileSync(
+      path.join(cliParentDir, 'package.json'),
+    )
     const { version } = JSON.parse(packageJsonBuffer)
 
     console.log(`sort-package-json ${version}`)
--- a/cli.js
+++ b/cli.js
@@ -6,7 +6,7 @@ import sortPackageJson from './index.js'
 const isCheckFlag = (argument) => argument === '--check' || argument === '-c'
 const isHelpFlag = (argument) => argument === '--help' || argument === '-h'
 const isVersionFlag = (argument) =>
-  argument === '--version' || argument === '-V'
+  argument === '--version' || argument === '-v'
 
 const cliArguments = process.argv.slice(2)
 const isCheck = cliArguments.some(isCheckFlag)
--- a/cli.js
+++ b/cli.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 import { globbySync } from 'globby'
 import fs from 'node:fs'
+import { fileURLToPath } from 'url'
 import sortPackageJson from './index.js'
 
 const isCheckFlag = (argument) => argument === '--check' || argument === '-c'
@@ -27,11 +28,10 @@ Strings passed as files are parsed as globs.
   )
   process.exit(0)
 }
-
 if (isVersion) {
   try {
     const cliParentDir = new URL('package.json', import.meta.url)
-    const packageJsonBuffer = fs.readFileSync(cliParentDir.pathname)
+    const packageJsonBuffer = fs.readFileSync(fileURLToPath(cliParentDir))
     const { version } = JSON.parse(packageJsonBuffer)
 
     console.log(`sort-package-json ${version}`)
--- a/cli.js
+++ b/cli.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 import { globbySync } from 'globby'
 import fs from 'node:fs'
+import path from 'node:path'
 import sortPackageJson from './index.js'
 
 const isCheckFlag = (argument) => argument === '--check' || argument === '-c'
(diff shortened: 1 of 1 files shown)


cli.js
```diff
@@ -1,12 +1,47 @@
...
+  `,
+  )
+  process.exit(0)
+}
+
+if (isVersion) {
+  try {
+    const cliRealPath = fs.realpathSync(process.argv[1])
+    const cliParentDir = (await import('node:path')).dirname(cliRealPath)
+    const packageJsonBuffer = fs.readFileSync(`${cliParentDir}/package.json`)
```
fisker, line 35:
```suggestion
    const packageJsonBuffer = fs.readFileSync('./package.json')
```

cli.js
```diff
@@ -1,12 +1,47 @@
...
+    const { version } = JSON.parse(packageJsonBuffer)
+
+    console.log(`sort-package-json ${version}`)
+  } catch {
+    // If anything goes wrong, fall back to 'unknown version'
```
fisker, line 40: I don't think this is possible to throw.

cli.js
aarondill: This suggestion won't work because the CLI can be run from any location, but I need to read the sort-package-json/package.json file specifically. Replacing it with the current directory (.) will result in either an error (if no package.json file is present) or reading another package's version number.

aarondill: I've been struggling with that very line for a while, and this is the best solution I could come up with.

cli.js
aarondill: I'm not sure exactly what circumstances might cause it to throw as I don't know the fs or path modules by heart, but file systems vary enough across OS's that I feel much better putting it in a catch block

aarondill:
Having just tested this revision on a windows OS, I can confirm that the help and version options work at the very least. It was not an extensive test, but I ran it with both options, ran it from a few different CWDs, and created a symlink and moved that around, and everything worked!

cli.js
aarondill: Having done some more research, I have greatly simplified the process of finding the project.json file using import.meta.url.

aarondill:
So, my most recent commit greatly improves the process of locating the project.json file for the version, however it doesn't work on windows because it produces a `/C:/FILEPATH` type file, while windows paths don't begin with /.
I can revert to the previous way, but that requires importing node:path and about 3 extra conversion steps.

keithamus (dismissed):
LGTM

cli.js
fisker: Sorry, you are right. I don't know what am I thinking. Using `import.meta.url` is the right way.

cli.js
```diff
@@ -1,12 +1,46 @@
...
+  -h, --help                 display this help and exit
+  -V, --version              display the version and exit
+  `,
+  )
+  process.exit(0)
+}
+if (isVersion) {
+  try {
+    const cliParentDir = new URL('package.json', import.meta.url)
+    const packageJsonBuffer = fs.readFileSync(fileURLToPath(cliParentDir))
```
fisker, line 34: We don't need call `fileURLToPath` here, `URL` can be used directly here.

cli.js
fisker: Let's remove the `try/catch`.

aarondill:
This should now be ready for review. The try-catch block has been removed, and I removed the extra call/import to `fileURLToPath`. I have tested this on both Ubuntu Bash and Windows 11 command line(through `node cli.js`) from many different CWDs and even through a symlink.

aarondill:
Reopening this PR according to @fisker's advice to merge this PR seperate to #283

cli.js
```diff
@@ -1,12 +1,40 @@
...
+  -c, --check                check if FILES are sorted
+  -q, --quiet                don't output success messages
+  -h, --help                 display this help and exit
+  -V, --version              display the version and exit
+  `,
+  )
+  process.exit(0)
+}
+if (isVersion) {
+  const cliParentDir = new URL('package.json', import.meta.url)
```
fisker, line 31: name it `packageJsonUrl`?

cli.js
```diff
@@ -1,12 +1,40 @@
...
+if (isHelp) {
+  console.log(
+    `Usage: sort-package-json [OPTION...] [FILE...]
+Sort npm package.json files. Default: ./package.json
+Strings passed as files are parsed as globs.
+
+  -c, --check                check if FILES are sorted
+  -q, --quiet                don't output success messages
+  -h, --help                 display this help and exit
+  -V, --version              display the version and exit
```
fisker, line 25: Why don't we use lowercased `-v`?

aarondill: The standard for cli tools and commands is that lowercase -v means verbose and uppercase -V means version

cli.js
aarondill: just did. that was a remnant from previous commits

cli.js
fisker: I don't think that's true, https://nodejs.org/api/cli.html#-v---version

aarondill: it seems there is not necessarily a standard on paper from my research, but in my usage of linux, ive generally noted -V to be version and -v to be verbose. if we ever plan to add a verbosity switch, we should leave version as -V, but if you still think it should be lowercase, i'll change it

fisker: From my experience, many tools uses `-v` for version. Let's take it, and not worry about `--verbose` for now.

cli.js
```diff
@@ -1,12 +1,40 @@
...
```
fisker, line 25:
```suggestion
  -v, --version              display the version and exit
```

cli.js
```diff
@@ -1,12 +1,40 @@
...
+const isVersion = cliArguments.some(isVersionFlag)
+
+if (isHelp) {
+  console.log(
+    `Usage: sort-package-json [OPTION...] [FILE...]
+Sort npm package.json files. Default: ./package.json
+Strings passed as files are parsed as globs.
+
+  -c, --check                check if FILES are sorted
+  -q, --quiet                don't output success messages
```
fisker, line 23: ``suggestion
```

Not implemented yet.

fisker (approved):
Thanks for working on this!

PR #283, opened 2023-01-22
refactor(cli): use `util.parseArgs` to parse arguments

### LINKED ISSUE #280: Quiet CLI option to oppress success output

aarondill:
This PR uses node:util.parseArgs for parsing arguments, allowing easy future extendability. The output of the tool remains the same as previous commits but user usage may change. Previous use cases which did not rely on the use of arguments beginning in '-' should remain unchanged, but in this update,  arguments which begin with a dash must follow the argument '--', which follows general user expectation. Each of the options(check, quiet, version, and help) have a corresponding short option(c,q,V,h) which can be combined into one (ie -cq = -c -q).


fisker:
Why don't we add #282 first and then do a refactor here after.

aarondill:
this should now be up to date with the changes in #282

README.md
```diff
@@ -69,6 +69,15 @@ $ sort-package-json "**/package.json" --check
 # 2 of 5 matched files are not sorted.
 ```
 
+#### `--quiet` flag
```
fisker, line 72: Better split this to anothet PR. It will be easier to merge a refactor  PR.

aarondill: I created #281, but I will need to edit it before reopening to ensure no merge conflicts. I can't do that right now, but hopefully I will have it done before the end of the day

cli.js
```diff
@@ -1,53 +1,129 @@
 #!/usr/bin/env node
-import fs from 'node:fs'
 import { globbySync } from 'globby'
+import fs from 'node:fs'
+import { parseArgs } from 'node:util'
```
fisker, line 4: I believe we still support Node.js v12, https://github.com/keithamus/sort-package-json/blob/dc1ad5a340d1b03ff35d39f267f0b7a827b1a06f/.github/workflows/pr.yml#L17 this is not available.

cli.js
```diff
@@ -1,53 +1,129 @@
...
 import sortPackageJson from './index.js'
 
-const isCheckFlag = (argument) => argument === '--check' || argument === '-c'
+const cliArguments = parseCliArguments()
+
+if (cliArguments.values.help) help()
+else if (cliArguments.values.version) version()
```
fisker, line 10: These two function name should start with verbs. 
Maybe `show` or `display`.

cli.js
aarondill: I would love a pointer on how to resolve this. Do I need to find a polyfill, or a different module, or is there some better way to include the module that works with older versions?

fisker: I'd suggest to postpone this change before we drop these old versions. It's refactor, not changing the program behavior.

aarondill: do we have a road map for when the old versions might be dropped?

fisker: It's common to follow the Node.js release schedule. https://github.com/nodejs/release#release-schedule

aarondill: if i'm reading this right, we're looking at October for 1.16.x to be discontinued? is that correct?

fisker: [`parseArgs`](
https://nodejs.org/api/util.html#utilparseargsconfig) is available in Node.js v 16, so 2023-04-30(EOL of v14) should be fine. 

Like I said it's a refactor so it can wait.

aarondill: alternatively, we *could* use this [parseargs polyfill](https://github.com/pkgjs/parseargs) to release this now and then change the import to the node:util version when 14 is discontinued

fisker: Yes we can, but I don't think worth to introduce a new dependency just for this.

Why don't you focus on adding `--quite` for now? Simple array filter should good enough.

fisker:
Let's do arg parse refactor here? We can merge later https://github.com/keithamus/sort-package-json/pull/283#discussion_r1083528211 .

aarondill:
Should I edit on top of master or #284?

fisker:
#284 don't touch arg parse part, you can start from master branch, changes should only effect L78-L110

aarondill:
> arg parsing should only effect L78-L110, isn't it?

That does seem correct

aarondill:
This has been refactored on top of the main branch. The node 12 and 14 tests have both failed, as expected, because parseArgs is not supported in those versions.

cli.js
```diff
@@ -27,6 +28,25 @@ If file/glob is omitted, './package.json' file will be processed.
...
+        version: { type: 'boolean', short: 'v' },
+        help: { type: 'boolean', short: 'h' },
+      },
+      allowPositionals: true,
+      strict: true,
+    })
+  } catch (err) {
+    const { message } = err
+    console.error(message)
+    process.exit(2)
```
fisker, line 46: We should move try/catch to `run`.

cli.js
```diff
@@ -58,35 +78,19 @@ function sortPackageJsonFiles(patterns, options) {
...
-      isCheck = true
-    } else if (argument === '--quiet' || argument === '-q') {
-      shouldBeQuiet = true
-    } else {
-      patterns.push(argument)
-    }
-  }
+  const patterns = cliArguments.positionals
+  const isCheck = !!cliArguments.values.check
+  const shouldBeQuiet = !!cliArguments.values.quiet
```
fisker, line 93: Shouldn't it already a boolean?

cli.js
```diff
@@ -27,6 +28,25 @@ If file/glob is omitted, './package.json' file will be processed.
   )
 }
 
+function parseCliArguments() {
+  try {
+    return parseArgs({
```
fisker, line 33: Just stylish, we can

````js
const {
	positionals: patterns,
	values: flags,
} = parseArgs({})

return {patterns, flags}
```

easier to use/understand.

fisker: Example I wrote before [1](https://github.com/prettier/prettier/blob/97c8e3c18ff3299c76e0d4c1a7374ebce0735f00/scripts/release/parse-arguments.js#L4-L12), [2](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/fcd893465f8e3cc59d34d6d1db533cea94247a0c/test/integration/test.mjs#L33-L37)

aarondill:
I have moved the try-catch to the run function, and moved destructuring to the parseCliArguments function. Also, there are now defaults(which I previously believed didn't work on node 16), so the boolean conversion is no longer needed. 
The parseCliArguments function returns an object with options and patterns. I could separate the options object into further properties, but that would force the seperated definition and assignment of several new variables, which I prefer to avoid

cli.js
```diff
@@ -58,41 +73,34 @@ function sortPackageJsonFiles(patterns, options) {
...
-    if (argument === '--check' || argument === '-c') {
-      isCheck = true
-    } else if (argument === '--quiet' || argument === '-q') {
-      shouldBeQuiet = true
-    } else {
-      patterns.push(argument)
-    }
-  }
-
   if (!patterns.length) {
```
fisker, line 96: Let's move this into `parseCliArguments`, it's like arguments normalization.

fisker:
>  there are now defaults

Nice! Good to know.

fisker:
Little help here, made a mistake in #284 

https://github.com/keithamus/sort-package-json/blob/07820dbf7af693b3d70aa1ad7d1cfde86cf254af/reporter.js#L63

Should be `shouldBeQuiet`

fisker:
Very nice! Can you add a test for `--unknown-flag`?

aarondill:
Is that check in reporter.js on line 73(/63) necessary? We already define the logger functions based on options.shouldBeQuiet on line 18.

fisker:
> Is that check in reporter.js on line 73(/63) necessary? 

It's not, this is why test still pass, but can reduce necessary calles.

fisker:
> Very nice! Can you add a test for `--unknown-flag`?

Also `--no-check` (it's common to parse as `check: false` in other lib)

aarondill:
> > Very nice! Can you add a test for `--unknown-flag`?
> 
> Also `--no-check` (it's common to parse as `check: false` in other lib)

Should this be a global modifier, or only the last counts? ie, should `sort-package-json --no-check --check` check or not?

fisker:
Better use `--no-version`? No one expect it to work.

aarondill:
> Better use `--no-version`? No one expect it to work.

You *do* want --no-version to be a legal argument?

fisker:
Sorry for not been clear, I didn't expect you to handle `--no-check`.

aarondill:
> Sorry for not been clear, I didn't expect you to handle `--no-check`.

Oh, so you just want a test that throws? Or do you prefer the new behavior?

fisker:
> You do want --no-version to be a legal argument?

No, I just want snapshot to show how `--no-version` works, not care about the result. Same for other `--no-*`, no need special treatment. Previous version is good.

fisker (approved):
Great work!

cli.js
```diff
@@ -27,6 +28,25 @@ If file/glob is omitted, './package.json' file will be processed.
   )
 }
 
+function parseCliArguments() {
+  const { values: options, positionals: patterns } = parseArgs({
+    options: {
+      check: { type: 'boolean', short: 'c', default: false },
+      quiet: { type: 'boolean', short: 'q', default: false },
+      version: { type: 'boolean', short: 'v', default: false },
+      help: { type: 'boolean', short: 'h', default: false },
```
fisker, line 37: [Docs says `default` only supported in Node.js v18](https://nodejs.org/api/util.html#utilparseargsconfig), since our flags all default to `false`, so it really matters.

aarondill: It says that, but running the program in node js v16.19.0, the default field works

fisker: Click `history` under title.

fisker: image

aarondill: I already saw that, but when run in node.js v16.19.0, the default fields *do* work. If you would like me to create manual defaults in parseCliArguments, I can do so; but in the current node.js v16 version, it works as it is.

fisker: Didn't test. Good to know. As I said it doesn't matters.

cli.js
```diff
@@ -58,41 +78,33 @@ function sortPackageJsonFiles(patterns, options) {
...
+  try {
+    ;({ options, patterns } = parseCliArguments())
+  } catch (error) {
+    process.exitCode = 2
+    console.error(error.message)
+    if (
+      error.code === 'ERR_PARSE_ARGS_UNKNOWN_OPTION' ||
+      error.code === 'ERR_PARSE_ARGS_INVALID_OPTION_VALUE'
+    ) {
+      console.error(`Try 'sort-package-json --help' for more information.`)
```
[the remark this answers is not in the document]
aarondill: I prefer the current mode because it results in shorter output, making it more immediately clear what went wrong, rather than the user having to find the error message above the wall of text. If you are insistent that we should show the help menu, perhaps we move the error displaying to after the help, so the user can see what went wrong without scrolling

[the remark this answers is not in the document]
aarondill: @keithamus, thoughts?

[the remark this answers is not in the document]
keithamus: I'm happy displaying an error message but I think it's important to show the help. Perhaps a more explicit error of `--foo is not a valid argument` would also be better here?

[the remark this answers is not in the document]
aarondill: Using the util.parseArgs function doesn't seem to expose a way to determine the invalid argument which was passed, other than parsing error.message; however error.message has no guarantee of consistency between versions ([Errors | Node.js](https://nodejs.dev/en/api/v18/errors/#errorcode)).

Regarding displaying help, would you suggest the following, or the error at the end? I feel that the help might be slightly overwhelming when just searching for the error.

```console
$ sort-package-json --not-an-option
Unknown option '--not-an-option'. To specify a positional argument starting with a '-',
place it at the end of the command after '--', as in '-- "--not-an-option"
Usage: sort-package-json [options] [file/glob ...]

Sort package.json files.
If file/glob is omitted, './package.json' file will be processed.

  -c, --check   Check if files are sorted
  -q, --quiet   Don't output success messages
  -h, --help    Display this help
  -v, --version Display the package version

```

I personally prefer the shorter output (see below), as it directly tells the user what went wrong, without having to find it, as well as telling them how to find help if they aren't sure already how to fix their error.
```console
$ sort-package-json --not-an-option
Unknown option '--not-an-option'. To specify a positional argument starting with a '-',
place it at the end of the command after '--', as in '-- "--not-an-option"
Try 'sort-package-json --help' for more information.
```

Additionally, this is the format followed by many default tools, such as `ls`, `df`, `du`, and `diff`:
```console
$ df --not-an-option
df: unrecognized option '--not-an-option'
Try 'df --help' for more information.
$ ls --not-an-option
ls: unrecognized option '--not-an-option'
Try 'ls --help' for more information.
$ du --not-an-option
du: unrecognized option '--not-an-option'
Try 'du --help' for more information.
$ diff --not-an-option
diff: unrecognized option '--not-an-option'
diff: Try 'diff --help' for more information.
```

The other reasonable possibility is to display an abbreviated usage menu, such as the one displayed by `git`:
```console
$ git --not-an-option
unknown option: --not-an-option
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]
```

aarondill:
> https://github.com/keithamus/sort-package-json/pull/283#discussion_r1083528211
are we ready to start looking at merging this PR again, because the node 12 EOL was reached, or are we still supporting V12?

fisker:
I think it's time to drop support for Node.js<=18. I'll work on it, and we can merge this after.

fisker (approved):

fisker (approved):

keithamus (approved):

PR #285, opened 2023-01-30
refactor: refactor cli

fisker:
- Separate features into different functions, so we can easily exit.
- Use `process.exitCode` instead of `process.exit()`
- Change exit code to 1 (file not sorted) and 2 (fatal errors)
- Update husky and lint-staged setup
- Minor tweak to the help information

---

Q: Should we consider breaking change for changing the exit code?


keithamus (approved):
I personally don't think error code changes are breaking, but I will leave it down to you to rebase commits as necessary.

cli.js
```diff
@@ -3,91 +3,113 @@ import { globbySync } from 'globby'
...
+  ) {
+    return showVersion()
+  }
 
-files.forEach((file) => {
-  const packageJson = fs.readFileSync(file, 'utf8')
-  const sorted = sortPackageJson(packageJson)
+  const patterns = []
+  let isCheck = false
+  let shouldBeQuit = false
```
aarondill, line 96: Every mention of `shouldBeQuit` should be spelled `shouldBeQuiet`

PR #286, opened 2023-01-30
refactor: make CLI async

fisker:
~This is based on #278, will rebase after #278 merged.~

Forgot we still support Node.js v12, we'll have to wait untill we drop it.


cli.js
```diff
@@ -57,33 +56,36 @@ function parseCliArguments() {
...
   for (const file of files) {
+    reporter.reportFound(file)
+
     try {
-      sortPackageJsonFile(file, reporter, isCheck)
+      await sortPackageJsonFile(file, reporter, isCheck)
     } catch (error) {
       reporter.reportFailed(file, error)
     }
   }
```
fisker, line 87: Still running serially, hard to make test pass if switch to parallel, I guess we can revist later.

cli.js
```diff
@@ -1,15 +1,14 @@
 #!/usr/bin/env node
-import fs from 'node:fs'
+import fs from 'node:fs/promises'
 import streamConsumers from 'node:stream/consumers'
 import { parseArgs } from 'node:util'
-import { globSync } from 'tinyglobby'
+import { glob } from 'tinyglobby'
 import sortPackageJson from './index.js'
 import Reporter from './reporter.js'
+import packageJson from './package.json' with { type: 'json' }
```
fisker, line 8: Should I move this to `showVersion`?

keithamus (approved):

PR #290, opened 2023-02-01
Update Usage in README

aarondill:
This updates the usage section in the readme to reflect the new output

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,10 @@ $ cat package.json
 }
 
 $ npx sort-package-json
-package.json is sorted!
+# package.json is sorted!
+#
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,7 +31,10 @@ $ cat package.json
 }
 
 $ npx sort-package-json
-package.json is sorted!
```
fisker, line 34: Let change the lang to console, so we don't need add `#` before output.

aarondill: So create a new code block with lang of console? is this a change we want to make though the entire readme?

keithamus (approved):

PR #291, opened 2023-02-02
chore: meta tweak

aarondill:
This fix returns the output of `npm run update-snapshot` to the closest to the previous output as is possible. The new increased output of cli.js takes from the focus of the script by displaying multiple lines at the end regardless of the state of package.json, so this PR silences the cli, instead sorting the package in the background, potentially even without user awareness.


package.json
```diff
@@ -34,7 +34,7 @@
     "sort-package-json": "node cli.js package.json --check",
     "test": "ava && dtslint --localTs node_modules/typescript/lib",
     "test-coverage": "nyc ava",
-    "update-snapshot": "ava -u && node cli.js package.json"
+    "update-snapshot": "ava -u && node cli.js package.json --quiet"
```
fisker, line 37:
```suggestion
    "update-snapshots": "ava --update-snapshots && node cli.js package.json --quiet"
```

fisker:
Maybe we should change

```diff
- node cli.js package.json
+ node cli.js \"package.json\"
```

To show user that better quote pattern/file s.

aarondill:
> Maybe we should change
> 
> ```diff
> - node cli.js package.json
> + node cli.js \"package.json\"
> ```
> 
> To show user that better quote pattern/file s.
> 
> Because shell may expand glob pattern and get unexpected result.

I could change it to `node cli.js 'package.json'` (single quotes avoids need to escape them) if this concerns you

fisker:
single quote doesn't work on windows.

fisker:
single quote doesn't work on windows.

.github/workflows/pr.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: npm run lint
 
       - name: Check package.json
-        run: npm run sort-package-json -- check
+        run: npm run sort-package-json -- --check
```
fisker, line 42: Nice catch.

NPM is making things weird, previously `--` is not needed, already run into this problem several times.

README.md
```diff
@@ -157,7 +157,7 @@ If an array, sort keys in ordering of `options.sortOrder`.
 
 ```js
 const sorted = sortPackageJson(packageJsonObject, {
-  sortOrder: ['version']
+  sortOrder: ['version'],
```
aarondill, line 160: Thank god for this! Having to disable prettier and undo the added commas every time I edited the readme was terrible

package.json
```diff
@@ -28,11 +28,10 @@
     "reporter.js"
   ],
   "scripts": {
-    "fix": "eslint . --fix && prettier . \"!**/*.js\" --write",
-    "lint": "eslint . && prettier . \"!**/*.js\" --check",
+    "fix": "eslint . --fix && prettier . \"!**/*.js\" --write && node cli.js \"package.json\"",
+    "lint": "eslint . && prettier . \"!**/*.js\" --check && node cli.js \"package.json\" --check",
     "prepare": "husky install",
     "semantic-release": "semantic-release",
-    "sort-package-json": "node cli.js \"package.json\"",
```
aarondill: Should we preserve this script for developer usage?

package.json
```diff
@@ -28,13 +28,13 @@
...
-    "lint": "eslint .",
+    "fix": "eslint . --fix && prettier . \"!**/*.js\" --write && node cli.js \"package.json\"",
+    "lint": "eslint . && prettier . \"!**/*.js\" --check && node cli.js \"package.json\" --check",
     "prepare": "husky install",
     "semantic-release": "semantic-release",
-    "sort-package-json": "node cli.js package.json --check",
     "test": "ava && dtslint --localTs node_modules/typescript/lib",
     "test-coverage": "nyc ava",
-    "update-snapshot": "ava -u && node cli.js package.json"
+    "update-snapshots": "ava --update-snapshots"
```
fisker, line 37: Let's remove it, we already got new "fix" script.

fisker (approved):

PR #303, opened 2023-10-03
Add bun entry

cupcakearmy:
Add entry for [bun](https://bun.sh/)

https://bun.sh/docs/runtime/modules#importing-packages

### DIFF
--- a/index.js
+++ b/index.js
@@ -214,6 +214,7 @@ const fields = [
   { key: 'exports' },
   { key: 'main' },
   { key: 'svelte' },
+  { key: 'bun' },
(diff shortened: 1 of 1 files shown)


keithamus (changes requested):
The documentation you linked to suggests that the `exports` key can contain `bun`, but there is no top level `bun` key in those docs, AFAICT. Can you point it out please?

Also this is missing tests.

cupcakearmy:
You are absolutely right, sorry for not being thorough enough. Will close as the sorting the `exports` field afaik alters behaviour, as the order is relevant for some tools.

PR #304, opened 2023-12-06
feat: Introduce option for sorting deps in the same ordering as npm

### LINKED ISSUE #234: Use same sorting logic for sorting dependencies as npm v7
NPM v7 changed the way dependencies are sorted very slightly which means this:

```
{
  "name": "spj-te",
  "dependencies": {
    "sort-package-json": "^1.52.0"
  },
  "devDependencies": {
    "@types/koa-bodyparser": "^4.3.3",
    "@types/koa__router": "^8.0.8"
  }
}
```

becomes this:

```
{
  "name": "spj-te",
  "dependencies": {
    "sort-package-json": "^1.52.0"
  },
  "devDependencies": {
    "@types/koa__router": "^8.0.8",
    "@types/koa-bodyparser": "^4.3.3"
  }
}
```

The specific change they made was to use `localeCompare`, which technically the better way for sorting strings than just calling `sort` but means npm v7 & above sort differently compared to both `sort-package-json` and the other two main package managers.

I opened an issue for discussion about this but it was closed with a comment about the change being intentional: https://github.com/npm/cli/issues/3935

I was thinking that `sort-package-json` could attempt to figure out if it should use `localeCompare` by trying to read `package-lock.json` and checking if the `lockfileVersion` is greater than 1, otherwise falling back to its current behaviour if its 1 or the lock cannot be read.

I'm happy to implement this change.

Nokel81:
- The option is `--npm-sort-ordering`, with the default value of `false`

resolves #234

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -29,9 +29,9 @@ If file/glob is omitted, './package.json' file will be processed.
   )
 }
 
-function sortPackageJsonFile(file, reporter, isCheck) {
+function sortPackageJsonFile(file, reporter, isCheck, sortUsingNpmV7Ording) {
   const original = fs.readFileSync(file, 'utf8')
-  const sorted = sortPackageJson(original)
+  const sorted = sortPackageJson(original, { sortUsingNpmV7Ording })
   if (sorted === original) {
     return reporter.reportNotChanged(file)
   }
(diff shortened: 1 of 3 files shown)


Nokel81:
I know that this is rather hacky, but I could not figure out another way to pass in an argument for the `over`'s

Also, how would I go about adding a unit test for this?

keithamus (changes requested):
Thanks for the PR! I’d like us to avoid the use of flags. We should be able to infer the npm version from the package.json or the users environment variables. Configuration flags mean that different invocations result in different outputs and I want to avoid that as much as possible.

keithamus:
I'm going to close this out, for a couple of reasons:

1. This hasn't seen updates since 2023, so I don't know if this is being actively pursued by the contributor.
2. It needs a rebase, so cannot be merged in it's current state.
3. Most importantly, I'd like to avoid the use of configuration via CLI, as mentioned in the above comment.

It's probably worth raising a new PR if you'd like to pursue an alternative that uses heuristics instead of CLI flags.

PR #305, opened 2023-12-14
Add docs for using sort-package-JSON via trunk

joshmarinacci:
[trunk check](https://docs.trunk.io/check) is an extendible superlinter with a built in language server and prexisting issue detection. sort-package-json is integrated [here](https://github.com/trunk-io/plugins/tree/main/linters/sort-package-json).

We're big fans of the sort-package-json project and hope our integration can increase its popularity!

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,44 @@ console.log(Object.keys(sorted))
 
 // -> [ 'dependencies', 'name', 'version' ]
 ```
+## Tools Integration
+
+### trunk check
+
+[trunk check](https://docs.trunk.io/docs) is an extendable superlinter with a builtin language 
+server and pre-existing issue detection. sort-package-json 
+is integrated [here](https://github.com/trunk-io/plugins/tree/main/linters/sort-package-json).
+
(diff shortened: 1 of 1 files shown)


keithamus (changes requested):
Thanks for the PR and I'm glad you're a big fan!

As is stands this PR takes up a large amount of the README and it seems like it's mostly to promote the trunk check product. I'm happy to see a link inside the `# Related tools` or `## Supported Libraries` subheadings but I think dedicating a whole new section to this seems a little much.

joshmarinacci:
Hi Keith. I did it this way to mirror some of the other tools we work with. For example **actionlint** has a whole section on Tools Integration that shows how to use it with reviewdog, precommit, VS code, and others in addition to Trunk Check.  However, if that's not what you want here, how about this:   

```
Trunk Check support (link) with:  `trunk check enable sort-package-josh`
```

https://github.com/rhysd/actionlint/blob/main/docs/usage.md

keithamus:
I'm going to close this one out, as it hasn't had activity since 2023, as well as my above commentary. At this point it might be worth raising a new PR if you wish to pursue such a change.

Issue #308, opened 2024-01-29
sorting scripts fails when switching from `npm-run-all` to `npm-run-all2`


viceice:
Renovate is now suggesting to replace the `npm-run-all` with `npm-run-all2` which is good, but on some repos `prettier` starts failing.
Any idea?

```diff
diff --git a/package.json b/package.json
index 9090468..58a787e 100644
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "author": "Michael Kriese <michael.kriese@visualon.de>",
   "type": "module",
   "scripts": {
-    "lint": "run-s prettier",
     "postinstall": "husky install",
+    "lint": "run-s prettier",
     "prettier": "prettier \"{**/*,*}.*\" --check --ignore-unknown",
     "prettier:fix": "prettier \"{**/*,*}.*\" --write --ignore-unknown"
   },
```

- https://github.com/containerbase/maven-prebuild/pull/108
- https://github.com/containerbase/base/pull/2044
- https://github.com/bcomnes/npm-run-all2/issues/129

viceice:
Seeing it on yarn v1 and pnpm v8, so seems unrelated to package manager

kachkaev:
Also seeing this in a PR created by Renovate bot. This should be fixed by https://github.com/keithamus/sort-package-json/pull/309.

PR #309, opened 2024-01-29
feat: don’t sort scripts when `npm-run-all2` is installed

kachkaev:
Closes https://github.com/keithamus/sort-package-json/issues/308. Context: https://github.com/keithamus/sort-package-json/pull/232

### DIFF
--- a/index.js
+++ b/index.js
@@ -163,7 +163,10 @@ const sortScripts = onObject((scripts, packageJson) => {
     return name
   })
 
-  if (!hasDevDependency('npm-run-all', packageJson)) {
+  if (
+    !hasDevDependency('npm-run-all', packageJson) &&
+    !hasDevDependency('npm-run-all2', packageJson)
+  ) {
     keys.sort()
   }
 
--- a/tests/scripts.js
+++ b/tests/scripts.js
@@ -68,3 +68,20 @@ for (const field of ['scripts', 'betterScripts']) {
     },
   })
 }
+
+for (const field of ['scripts', 'betterScripts']) {
+  test(`${field} when npm-run-all2 is not a dev dependency`, macro.sortObject, {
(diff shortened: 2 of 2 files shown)


keithamus (approved):
lol

bcomnes:
You can also add numerics to your script names to force a specific sort order while retaining the ability to alphanumeric sort.

viceice:
> You can also add numerics to your script names to force a specific sort order while retaining the ability to alphanumeric sort. IMO this is preferable than adding tool specific behaviors. 

doesn't work for `postinstall`. I would expect to have scripts sorted alphanumeric, but `postinstall` was sorted before `lint`. 🤔

benquarmby:
If I wanted to control the order of execution with a tool like `npm-run-all2`, why the flerking shnit wouldn't I be explicit about it?:
```diff
{
-    "do": "run-s do:*"
+    "do": "run-s do:clean do:build do:translate do:version do:publish"
}
```
Relying on a magic `*` and the arbitrary order of scripts is fragile.

Excuse the rant, but I'm frustrated that I can no longer use `npm-run-all2` as an escape hatch from this silly "skip sorting scripts when npm-run-all is installed" behavior. Scripts should always be sorted! This behavior is promoting bad practice.

bcomnes:
I generally agree that disabling behavior automatically as a side effect of other tools being detected is an odd design choice. If order of execution matters in your npm-run-all scripts, you should add numerics to your scripts to guarantee order, even when external tools run stylistic sorts on them. This approach lets you keep your top level script short, but still define explicit order.

```json
{
"do": "run-s do:*"
"do:01-clean": "clean"
"do:02-build": "build"
"do:03-translate": "translate"
"do:03-version": "version"
"do:03-publish": "publish"
}
```

PR #313, opened 2024-03-26
feat: add pnpm

samsternatretool:
`pnpm` (https://pnpm.io/) is a popular package manager that has support for advanced configuration via `package.json` fields. This keeps those deep keys sorted.


tests/fields.js
```diff
@@ -258,3 +258,59 @@ test('badges', (t) => {
     'Should sort `badges[]`',
   )
 })
+
+test('pnpm', macro.sortObject, {
```
samsternatretool, line 262: If this test looks overly specific it's because I took a real world example from my own repository (nothing private here) and set the expectation of how we'd like it sorted.

samsternatretool:
@keithamus do you want me to delete that test?

keithamus (approved):
All good. I was mistaken about the validity of the test.

PR #314, opened 2024-03-27
feat: improve pnpm sorting

westrik:
https://github.com/keithamus/sort-package-json/pull/313 added the ability to deeply sort a `pnpm` object in `package.json`. This PR improves `pnpm` sorting in the following ways:

- groups similar fields together
- sorts `pnpm.overrides` entries with the same package name using semver rather than alphabetically

**Example:**

Before this change:
```
"pnpm": {
  "overrides": {
      "@react-stately/selection@~3.10.3": "3.10.3",
      "@react-stately/selection@~3.7.0": "3.7.0"
  }
}
```

After this change:
```
"pnpm": {
  "overrides": {
      "@react-stately/selection@~3.7.0": "3.7.0",
      "@react-stately/selection@~3.10.3": "3.10.3"
  }
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,21 +57,31 @@ const sortGitHooks = sortObjectBy(gitHooks)
 
 const sortObjectBySemver = sortObjectBy((a, b) => {
   const parseNameAndVersionRange = (specifier) => {
-    const atMatches = [...specifier.matchAll('@')]
+    const [nameAndVersion, childName] = specifier.split('>')
+    const atMatches = [...nameAndVersion.matchAll('@')]
     if (
       !atMatches.length ||
       (atMatches.length === 1 && atMatches[0].index === 0)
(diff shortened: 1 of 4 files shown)


keithamus:
@samsternatretool what do you think?

samsternatretool:
@keithamus LGTM from me! @westrik is on my team at Retool and he noticed some subtlety when it comes to sorting `pnpm` overrides after we merged my original change.

keithamus:
Some failing tests in Node 12 because of `Array#at`. Will need to use an alternative or polyfill.

westrik:
> Some failing tests in Node 12 because of `Array#at`. Will need to use an alternative or polyfill.

updated with alternative

keithamus (approved):

Issue #315, opened 2024-03-29
add dual release support


unional:
Hi @keithamus,

Since `sort-package-json` converted to pure ESM package, [VSCode sort-package-json extension](https://github.com/unional/vscode-sort-package-json) couldn't use the newer versions since electron does not support ESM.
I can help to add dual release to the package. 😉

unional:
Um... the `git-hooks-list` is a problem. It is using `import.meta.url`.

PR #316, opened 2024-03-29
Add dual release support

unional:
fix: #315


keithamus (approved):

package.cjs.json
```diff
@@ -0,0 +1 @@
+{ "type": "commonjs" }
```
fisker, line 1: @unional Can't we bundle file as `cjs/index.cjs`? So we can get ride of this file?

PR #325, opened 2024-10-12
Document check/diff example for stdio usage

imme-emosol:


README.md
```diff
@@ -97,6 +97,12 @@ To read from `stdin` and output the result to `stdout` use the `--stdin` flag.
 $ cat package.json | sort-package-json --stdin
 ```
 
+This can, for instance, be used to generate a diff before changing `package.json`.
+
+```bash
+$ ( PKG="./package.json" ; cat "${PGK?}" | pnpm sort-package-json --stdin | diff "${PKG?}" - ; )
```
keithamus, line 103:
```suggestion
$ ( PKG="./package.json" ; cat "${PGK?}" | sort-package-json --stdin | diff "${PKG?}" - ; )
```

keithamus (approved):
Nice example! Can we please remove the reference to `pnpm` just to make this a little easier for folks to use?

README.md
imme-emosol: will fix the typo too, while i'm at it (PGK instead of PKG :stuck_out_tongue_closed_eyes: )

keithamus (approved):

PR #326, opened 2024-10-15
chore: switch to tinyglobby

benmccann:
https://npmgraph.js.org/?q=globby - 23 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies


keithamus (approved):
LGTM. Looks as though tinyglobby is a fair bit faster to boot. Hopefully it works well.

PR #327, opened 2024-11-15
fix: use locale-independent comparison for sorting of overrides

ecl1ps:
The goal of this package is to format the package.json files always in the same way. But the `sortObjectBySemver` function uses locale comparison for package names. Some locales have different sorting rules so the result can be different on other machines with various locales. 

For example, the Czech locale 'cs' puts 'ch' after the 'g' letter. 

We should sort using locale-independent string sorting.


index.js
```diff
@@ -77,7 +77,7 @@ const sortObjectBySemver = sortObjectBy((a, b) => {
   const { name: bName, range: bRange } = parseNameAndVersionRange(b)
 
   if (aName !== bName) {
-    return aName.localeCompare(bName)
+    return aName > bName ? 1 : -1
```
keithamus, line 80: The change to use `localeCompare` came from https://github.com/keithamus/sort-package-json/pull/314 and seems well motivated. I understand that it's going to be dependant on the locale, which isn't great, but moving back to comparing character points seems sub-optimal. I wonder if fixing the locale would be a better choice?

```suggestion
    return aName.localeCompare(bName, 'en')
```

ecl1ps: This was my first idea, but I didn't want to hardcode one specific language. But it will work. I am fine with it, English is de facto the standard language in the programming world anyway.

keithamus: I agree that it feels bad to hard-code a particular language but in this instance it seems well justified. no?

ecl1ps: Yeah, I agree. I applied your suggestion.

PR #329, opened 2024-11-23
feat: added support for ignore pattern to cli

### LINKED ISSUE #328: Performance affected change
this change affects the perormance as it is not ignoring node_modules by default and there is no option to exclude it
https://github.com/keithamus/sort-package-json/compare/v2.10.1...v2.10.2

spamshaker:
- defaults to node_modules

resolves #328


cli.js
```diff
@@ -43,28 +44,40 @@ function sortPackageJsonFile(file, reporter, isCheck) {
...
-
+  performance.mark('finish')
   reporter.printSummary()
+  if (process.env.DEBUG) {
+    console.log([
+      performance.measure('Glob Search', 'start', 'sorting'),
+      performance.measure('Sorting', 'sorting', 'finish'),
+      performance.measure('Total', 'start', 'finish'),
+    ])
+  }
```
keithamus, line 69: Can we remove this one please? I'd prefer to avoid shipping debug code. It's easy enough to add in for those interested in debugging.

spamshaker: sure

keithamus (approved):

PR #331, opened 2025-01-15
feat: add support for vscode l10n

chouchouji:
# Link

The internationalization of vscode 

image

[https://code.visualstudio.com/api/references/vscode-api#l10n](https://code.visualstudio.com/api/references/vscode-api#l10n)

[https://github.com/microsoft/vscode-extension-samples/blob/main/l10n-sample/package.json#L21](https://github.com/microsoft/vscode-extension-samples/blob/main/l10n-sample/package.json#L21)


chouchouji:
@keithamus I create a pr, please review it when you have time. Thanks.

keithamus (approved):

fisker:
Sorry that I'm late, but why it's not listed 
here?  https://code.visualstudio.com/api/references/extension-manifest

chouchouji:
> Sorry that I'm late, but why it's not listed here? https://code.visualstudio.com/api/references/extension-manifest

I just feel so weird too. They supports it, but don't update docs. The older version of vscode extension doesn't have this field, you can get it about vscode internationalization from this [commit](https://github.com/microsoft/vscode-extension-samples/commit/45531b256a5b39955c8ccff18bc9f8230b406820)

Issue #332, opened 2025-01-20
Another bug wrt sort pre/post scripts with colon together


rfalke-rtl:
Similar to https://github.com/keithamus/sort-package-json/issues/307

Sorted input with extra blank lines for clarification:
```
  "scripts": {
    "test": "echo",
    "test:coverage": "echo",

    "pretest:es-check": "echo",
    "test:es-check": "echo",
    "posttest:es-check": "echo",

    "test:types": "echo"
  },

```

Output with `npx sort-package-json@2.14.0`:

```
  "scripts": {
    "pretest:es-check": "echo",
    "test": "echo",
    "test:coverage": "echo",
    "test:es-check": "echo",
    "test:types": "echo",
    "posttest:es-check": "echo"
  }
```

keithamus:
PRs welcome!

rfalke-rtl:
After looking into this further is shows that 2.12.0 works as I would expect it. The change of https://github.com/keithamus/sort-package-json/pull/330 changed the behavior.

The question is what is the correct/intended behavior. I think the expected order in

```
{
  "scripts": {
    "prebuild": "run-s prebuild:*",
    "prebuild:1": "node prebuild.js 1",
    "prebuild:2": "node prebuild.js 2",
    "prebuild:3": "node prebuild.js 3",
    "build": "run-s build:*",
    "build:bar": "node bar.js",
    "build:baz": "node baz.js",
    "build:foo": "node foo.js",
    "postbuild": "run-s prebuild:*",
    "postbuild:1": "node prebuild.js 1",
    "postbuild:2": "node prebuild.js 2",
    "postbuild:3": "node prebuild.js 3",
    "d-unrelated": "..",
    "e-unrelated": "..",
    "f-unrelated": "..",
  }
}
```

is quite special and a generic tool like sort-package-json should support the common case (see my expectations from above) out of the box.

rfalke-rtl:
Ok than I propose https://github.com/keithamus/sort-package-json/pull/333 to revert the recent changes.

chouchouji:
> [@chouchouji](https://github.com/chouchouji) thoughts?

I check the issue and pr. The result of sorting pre/post scripts you expected is quite different with [https://github.com/keithamus/sort-package-json/issues/307](https://github.com/keithamus/sort-package-json/issues/307). I am not sure that which is better? Maybe we can take a discussion.

fisker:
Fixed by #333

PR #330, opened 2025-01-11
feat: sort pre/post scripts with colon together

chouchouji:
# Issue

close #307 

# Test

image

PR #333, opened 2025-01-21
feat: revert sort pre/post scripts with colon together (#332)

rfalke-rtl:
I'm not sure if the commit message is ok.


fisker (approved):
I agree we should revert, https://github.com/keithamus/sort-package-json/pull/330#issuecomment-2608518482 looks really bad.

PR #338, opened 2025-02-28
chore: update husky to v9

fisker:


keithamus (commented):
Looks like this fails on our older builds so we'll have to wait until we drop support here.

Having said that; maybe it's worth simply removing husky? We have CI which will run tests & lint, and we squash PRs into 1 commit, so I am not sure there's much value in doing it during commit.

fisker:
Make sense to remove the format and ESLint fix on commit.
But after a second thought, it also runs commitlint for commit messages, maybe worth to keep it?

keithamus (approved):

PR #344, opened 2025-03-01
chore: drop support for Node.js<20

fisker:
Update all deps except these already have separate PRs.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,10 +14,6 @@ jobs:
         node_version:
           - '22'
           - '20'
-          - '18'
-          - '16'
-          - '14'
-          - '12'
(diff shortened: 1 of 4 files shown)


keithamus:
This needs rebasing now. Should we wait for https://github.com/keithamus/sort-package-json/pull/336 first? I am ambivalent.

fisker:
IDK, #336 is currently doing more than fixing https://github.com/keithamus/sort-package-json/issues/294,  I think it's too risky. I hope you can post your opinion there.

fisker:
I just realized that we are currently not sorting `exports` at all  https://github.com/keithamus/sort-package-json/blob/43ee7e7f613d113e0621d78195158cfe1dd9df1f/index.js#L291, but somehow I thought #336 is fixing bugs... I guess we don't need to wait for it.

keithamus (approved):

PR #296, opened 2023-04-19
feat: add --stdin flag

danielpza:
Passing the `--stdin` flag would make it so it reads from `stdin` and outputs to `stdout`. This is useful for some tools/editors integration

Usage example:

```sh
$ cat package.json | sort-package-json --stdin
```

Closes https://github.com/keithamus/sort-package-json/issues/293

PR #345, opened 2025-03-01
refactor: replace `get-stdin` with native alternative

fisker:
This was mentioned in https://github.com/keithamus/sort-package-json/pull/296#discussion_r1171145098, but things get much easier now, after we drop support for old versions.


fisker:
Ah, bad merge.

fisker:
Good to go.

keithamus (approved):
LGTM

PR #350, opened 2025-03-04
feat:  sort export/import paths

GeorgeTaveras1231:
Subset of #336 

Sorts path objects in `imports` and `exports` alphabetically, pushing wildcards to bottom.


GeorgeTaveras1231:
@fisker this only handles the path keys - I'll let you handle the conditions.

fisker:
I just learned that conditions can be an array ... https://github.com/acornjs/acorn/blob/0d046aa11da7a8e545eda830037675f75d4d88bc/acorn/package.json#L9-L16

keithamus:
@GeorgeTaveras1231 would you kindly rebase this, and we can look to merge.

Issue #355, opened 2025-04-24
Sorting of package names with capital letters doesn't match npm


karlhorky:
Hi there, first of all thanks for this package, it's great! Especially great via [the "Sort package.json" VS Code extension](https://marketplace.visualstudio.com/items/?itemName=unional.vscode-sort-package-json).

I ran into an issue today where the sorting of packages starting with capital letters and packages starting with lowercase letters doesn't match in `sort-package-json` and npm:

1. `sort-package-json`: Sorts all packages starting with capital letters to the beginning, before all packages starting with lowercase letters
2. npm: Sorts package names alphabetically, ignoring uppercase or lowercase

Reproduction:

```bash
➜  p mkdir r
➜  p cd r
➜  r npm add axios JSONStream

added 26 packages in 818ms

6 packages are looking for funding
  run `npm fund` for details
➜  r cat package.json
{
  "dependencies": {
    "axios": "^1.9.0",
    "JSONStream": "^1.3.5"
  }
}
➜  r npx sort-package-json
package.json is sorted!

Found 1 file.
1 file successfully sorted.
➜  r cat package.json
{
  "dependencies": {
    "JSONStream": "^1.3.5",
    "axios": "^1.9.0"
  }
}
➜  r git init
Initialized empty Git repository in /Users/k/p/r/.git/
➜  r git:(main) ✗ echo 'node_modules' > .gitignore
➜  r git:(main) ✗ git add .
➜  r git:(main) ✗ git commit -m 'Sort capital letters first (sort-package-json)'
[main (root-commit) d0d7fde] Sort capital letters first (sort-package-json)
 3 files changed, 329 insertions(+)
 create mode 100644 .gitignore
 create mode 100644 package-lock.json
 create mode 100644 package.json
```

But then, adding a package using npm causes the `JSONStream` package to be moved beneath `json-schema` again:

```bash
➜  r git:(main) npm add json-schema

added 1 package, and audited 28 packages in 882ms

6 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
➜  r git:(main) ✗ git --no-pager diff package.json
diff --git a/package.json b/package.json
index c1d7d80..605a208 100644
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "dependencies": {
-    "JSONStream": "^1.3.5",
-    "axios": "^1.9.0"
+    "axios": "^1.9.0",
+    "json-schema": "^0.4.0",
[...]

PR #358, opened 2025-05-03
Sort dependencies with npm algorithm

### LINKED ISSUE #234: Use same sorting logic for sorting dependencies as npm v7

karlhorky:
Closes #234 #355

As per @keithamus' comment in #355, sorting via the npm CLI's algorithm would resolve #234 and #355

Not sure if this is the preferred way of adding this change - happy to continue this to make this ready to accept

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,10 @@ const sortObjectByIdent = (a, b) => {
   return 0
 }
 
+// sort deps like the npm CLI does (via the package @npmcli/package-json)
+// https://github.com/npm/package-json/blob/b6465f44c727d6513db6898c7cbe41dd355cebe8/lib/update-dependencies.js#L8-L21
+const sortDependenciesLikeNpm = sortObjectBy((a, b) => a.localeCompare(b, 'en'))
+
 // https://github.com/eslint/eslint/blob/acc0e47572a9390292b4e313b4a4bf360d236358/conf/config-schema.js
 const eslintBaseConfigProperties = [
   // `files` and `excludedFiles` are only on `overrides[]`
@@ -366,13 +370,14 @@ const fields = [
   { key: 'tap', over: sortObject },
   { key: 'oclif', over: sortObjectBy(undefined, true) },
   { key: 'resolutions', over: sortObject },
-  { key: 'dependencies', over: sortObject },
-  { key: 'devDependencies', over: sortObject },
+  { key: 'overrides', over: sortDependenciesLikeNpm },
+  { key: 'dependencies', over: sortDependenciesLikeNpm },
+  { key: 'devDependencies', over: sortDependenciesLikeNpm },
   { key: 'dependenciesMeta', over: sortObjectBy(sortObjectByIdent, true) },
-  { key: 'peerDependencies', over: sortObject },
+  { key: 'peerDependencies', over: sortDependenciesLikeNpm },
   // TODO: only sort depth = 2
   { key: 'peerDependenciesMeta', over: sortObjectBy(undefined, true) },
-  { key: 'optionalDependencies', over: sortObject },
+  { key: 'optionalDependencies', over: sortDependenciesLikeNpm },
   { key: 'bundledDependencies', over: uniqAndSortArray },
   { key: 'bundleDependencies', over: uniqAndSortArray },
   /* vscode */ { key: 'extensionPack', over: uniqAndSortArray },


index.js
```diff
@@ -366,13 +370,14 @@ const fields = [
   { key: 'tap', over: sortObject },
   { key: 'oclif', over: sortObjectBy(undefined, true) },
   { key: 'resolutions', over: sortObject },
-  { key: 'dependencies', over: sortObject },
-  { key: 'devDependencies', over: sortObject },
+  { key: 'overrides', over: sortDependenciesLikeNpm },
```
karlhorky, line 373: also introduces sorting for [the `'overrides'` key for npm Overrides](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides)

G-Rath:
This looks good to me, though you should update/add some tests ideally and also flagging I personally would prefer if this tried to check what package manager was being used since I think that'll be very cheap and the dependency objects are the only ones package managers actually sort themselves.

I'm happy to add that myself as a follow-up but want to mention it here all the same to confirm people are agreeable to that

karlhorky:
> you should update/add some tests ideally

Good idea, I'll take a look at this.

> I personally would prefer if this tried to check what package manager was being used

Hmm... although I can follow the idea behind it, I think I would have reservations with that:

1. That sorting change was made in npm CLI at least 4 years ago
2. The `localeCompare` sorting makes more sense than the current sorting (check my issue #355 for how package names starting with capital letters are sorted before all lowercase package names, which seems unusual to me on a human level)

G-Rath:
> I think I would have reservations with that:

NPM isn't the only package manager though, and it's the only one that does this method of sorting - that frustrates me to no end because it feels like a completely unneeded difference in our ecosystem, but when I tried talking about that with NPM they said [they were not going to change back because (ironically) they felt it would be too much churn for NPM v7+ users](https://github.com/npm/package-json/pull/2#issuecomment-1012267763), and I just don't see it being something that every really changes.

I also would love to use the same package manager everywhere, but that too is something I don't think will ever be "solved" as the big four have their own merits and fans meaning you probably shouldn't rule out using a particular one indefinitely - I frequently have to jump between the original three due to my open source work.

The reason why I care for it respecting the package manager is that this a tool that often only needs to get run on occasion as the majority of content in `package.json` are dependencies which as mentioned package managers handle sorting, so after a first-run there's not a lot of value in running this in CI every single time, so we just do so manually as and when needed - and most of the time that's fine as you just run the tool and commit the result, but every so often you get a codebase with a dependency whose sort order will be different and suddenly the diff has this weird package swapping and you've got to explain that to the junior dev and so on.

Hence my desire to try and be what I consider a little more friendly by doing a couple of cheap attempts to guess what package manager is being used and thus what sort order, such as `fs.stat('package-lock.json')` or maybe looking for `engines.npm`.

(fwiw if you do ever want to talk to any of the other package managers about changing their order, feel free to tag me for my vote - I don't actually think they'd be opposed to changing it just requires someone being able to put the time in to get it across the line since it'll be hard to make a high priority)

karlhorky:
Hm ok understood.

What do you think about a different way: an option `disableDepSorting` (default `false`) that would leave `dependencies`, `devDependencies`, etc unsorted for those who want that? That could be used by users of those other package managers but also by other users who don't want dependencies to be sorted... 🤔

karlhorky:
> The CLI purposely does not support configuration so that's a no-go

I see that now, in the [no-configuration section in the readme](https://github.com/keithamus/sort-package-json#i-would-like-this-tool-to-be-configurable-with-a-config-file-or-command-line-arguments):

> The lack of configuration here is a feature, not a bug. The intent of this tool is that a user can open a package json and always expect to see keys in a particular order.

However, reading this I also think that introducing any branching in the dep-sorting behavior based on package manager would also be against this same philosophy (because it would be "keys in a different order").

I have no horse in the race here - so I'm happy to go with whatever the maintainer team here decides. But I would suggest getting rid of the current dependencies sorting, because the uppercase-before-lowercase sorting doesn't make sense the way it is currently.

@keithamus any opinion here? Keep the tool consistently sorting regardless of package manager? Or introduce some "magic" detection for non-npm and disable dependency sorting for those?

G-Rath:
> However, reading this I also think that introducing any branching in the dep-sorting behavior based on package manager would also be against this same philosophy (because it would be "keys in a different order").

No because the sort would still be predictable and deterministic, like how the sorting of `scripts` is different depending on if dependencies like `npm-run-all` are present and if its `--parallel` flag is being used in scripts

keithamus (approved):
Great work! Let’s merge this as is, and we can iterate further if people consider this change too problematic. 

I’d like to avoid too many heuristics as with too many in place it can lead to a “sense” of indeterminism.

PR #361, opened 2025-05-08
fix: prevent sorting when npm-run-all2 is used

### LINKED ISSUE #359: bug: scripts are getting sorted, even with run-s included
In https://github.com/keithamus/sort-package-json/commit/9f1101b0dc82276c0b72052bac2117ae570b5be4 there was a change that now sorts scripts.

We are using npm-run-all2 and a run-s inside a script but not at the start

https://github.com/prettier/plugin-pug/blob/25329d1968f7aad2a9793f1ffe6658cd9f35518a/package.json#L19

however there is even a script with run-s at the start https://github.com/prettier/plugin-pug/blob/25329d1968f7aad2a9793f1ffe6658cd9f35518a/package.json#L9

and it still sorts them as it wants

is there an option I can set to disable script sorting?

ikemo3:
Add support for npm-run-all2 to prevent script sorting when sequential wildcard execution is needed.

Closes #359

### DIFF
--- a/index.js
+++ b/index.js
@@ -223,13 +223,16 @@ const hasDevDependency = (dependency, packageJson) => {
 }
 
 const runSRegExp =
-  /(?<=^|[\s&;<>|(])(?:run-s|npm-run-all .*(?:--sequential|--serial|-s))(?=$|[\s&;<>|)])/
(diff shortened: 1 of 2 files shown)


keithamus (approved):
Thanks for this @ikemo3!

PR #362, opened 2025-05-26
feat: ignore scripts option

Shinigami92:


Shinigami92:
It was by far more easier for me to contribute and switch to https://github.com/un-ts/prettier/tree/master/packages/pkg / https://www.npmjs.com/package/prettier-plugin-pkg

@JounQin also already released a v0.21.0 for me which includes this feature from now on 🚀

PR #364, opened 2025-06-07
chore: migrate type tests to TSTyche

### LINKED ISSUE #351: Replace `dtslint`
[`dtslint` is deprecated](https://www.npmjs.com/package/dtslint), should find a replacement, since I'm not good at TypeScript, open this and wait for others to contribute.

mrazauskas:
Closes #351

As it is mentioned in the issue, `dtslint` is deprecated. I would like to suggested using TSTyche instead. Must be mentioned that I am its author.

---

[`tstyche`](https://github.com/tstyche/tstyche) is built to work with projects of any scale. For example, currently in the [`effect` repo](https://github.com/Effect-TS/effect) it is running more than 3500 assertions against 5 versions of TypeScript.

It can bring you far, but TSTyche also plays nicely with small projects like `sort-package-json`:

- the install size is only 250kB (it loads currently installed TypeScript for testing);
- it has `test()` and `describe()` helpers to keep tests organized;
- and assertions like `.not.toBeCallableWith()` (none of other type testing libraries has this assertion).

Worth to note that the `index.d.ts` file also gets checked. Try adding some errors there and you will see them.

---

Happy to answer any questions or provide more details. You can also check the documentation website: https://tstyche.org


fisker (approved):
Thanks!

mrazauskas:
@fisker To draw your attention, the type tests currently are not running in the CI pipeline. If I don’t mix up anything, only the `test-coverage` script is invoked in GitHub Actions, but it does not include type testing:

https://github.com/keithamus/sort-package-json/blob/aa6774ad937feb83178c8bc981f08305e1d22b5c/package.json#L50

Perhaps it would make sense adding a separate `test-types` script?

fisker:
Thanks for pointing it out, using `c8 npm run test` now.

tsconfig.json
```diff
@@ -1,16 +1,13 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "lib": ["es6"],
+    "module": "node16",
+    "moduleResolution": "node16",
```
mrazauskas, line 4: Sorry, one more detail. Since the lowest supported Node.js version is 20, having module resolution set like this will handle self-referencing. Which is importing `sort-package-json` directly without `paths` (deleted below).

fisker: > using ESM syntax

We also bundle the CommonJS version, should we test it? https://github.com/keithamus/sort-package-json/blob/aa6774ad937feb83178c8bc981f08305e1d22b5c/package.json#L27

fisker: `publint` seems unhappy with the types. https://publint.dev/sort-package-json@3.2.1

mrazauskas: Ah.. This is out of scope for TSTyche. It only checks correctness types.

`publint` or `@arethetypeswrong/cli` can help with checking `package.json`. They both are unhappy. Here is result from `attw`: https://arethetypeswrong.github.io/?p=sort-package-json%403.2.1

By the way, `publint` recommends using both of these tools together ([source](https://publint.dev/docs/comparisons#are-the-types-wrong)). That sounds too much. Or? What you think?

I can also revert the `tsconfig.json` and this PR can land. The rest can be solved separately.

fisker (approved):

package.json
```diff
@@ -102,6 +101,6 @@
     "prettier": "^3.5.2",
     "semantic-release": "^24.2.3",
     "tempy": "^3.1.0",
-    "typescript": "^5.8.2"
```
mrazauskas, line 105: I realised that `typescript` and `tsconfig.json` are only used by `dtslint`, because of this line: https://github.com/keithamus/sort-package-json/blob/1c142c61f4cd370fc3b17f150927ed4d82efa6f3/package.json#L49

TSTyche does not require TypeScript to be installed. And also it sets strict and modern default compiler options: https://tstyche.org/project/compiler-options#default-compiler-options

PR #366, opened 2025-06-07
chore: remove `ncp`

mrazauskas:
This PR removes `ncp` that seems to be unused:

- it was added in #316
- became unused in #346

The `build` script is changing between these two PRs. Looks like that was the only usage of `ncp`. Unless I missed something.


fisker:
Thanks!

fisker (approved):

PR #369, opened 2025-06-23
feat: support workspaces

### LINKED ISSUE #368: does not alphabetically sort catalog
pnpm and Bun have a feature called "catalogs": https://bun.sh/docs/install/catalogs

In short, it is a field similar to "dependencies", so it needs to be alphabetized by this plugin. However, the plugin does not currently support this, so we need to add this feature.

Here's an example of a package.json file with a "catalog":

```json
{
  "name": "foo",
  "workspaces": {
    "packages": [
      "packages/*",
    ],
    "catalog": {
      "@types/bun": "1.2.17",
      "@types/ssh2-sftp-client": "9.0.4",
      "applicationinsights": "3.7.0",
      "complete-common": "2.3.0"
    }
  }
}
```

Zamiell:
Fixes #368.


tests/fields.js
```diff
@@ -32,7 +32,6 @@ for (const field of [
   'examplestyle',
   'assets',
   'man',
-  'workspaces',
```
fisker, line 35: Can we keep this?

Zamiell:
@victor-xplore That's interesting. I have not thought about this very deeply but my gut reaction is:
If you have workspaces that need to be tested in a very specific order, then representing that order inside the package.json file seems confusing and brittle. People onboarding to your monorepo will likely not realize that the ordering of your workspaces have magic behavior. Why not define a testing script that establishes the ordering explicitly, with clear documentation?

victor-xplore:
@Zamiell If the packages depend on each other, and needs to be complied, then you need to compile them in a specific order. As far as I know this is the only way to define what order npm will execute in when you use --workspaces. Introducing a script will mean one more thing to maintain, but I agree that relying on a less known functionality of npm can seem confusing, but it's not a complicated one. I guess another option is to introduce a lib like lerna, that takes into account dependencies when running scripts, but doing that only to fix something the linting breaks is not ideal.

> My instinct is that most users will want the workspaces alphabetically sorted for the same reasons that they will want the dependencies array sorted.
You might be right, I guess I'll have to stay on 3.2.2 or write that script, seems like a small thing to introduce a flag for.

PR #382, opened 2025-12-03
feat: sort dependencies according to detected package manager, support `devEngines` field

### LINKED ISSUE #363: Sorting algorithm fights yarn
The latest release broke CI for me. The order is correct and consistent on my local machine. But for some reason, when I run `prettier --fix` on CI it changes the sort order. This is what I get on CI:

```
index 3e57cb9..6a912df 100644
--- a/packages/babel-config/package.json
+++ b/packages/babel-config/package.json
@@ -46,9 +46,9 @@
   },
   "devDependencies": {
     "@cedarjs/framework-tools": "workspace:*",
+    "@types/babel-plugin-tester": "9.0.10",
     "@types/babel__core": "7.20.5",
     "@types/babel__register": "7.17.3",
-    "@types/babel-plugin-tester": "9.0.10",
     "@types/node": "20.17.10",
     "babel-plugin-tester": "11.0.4",
     "tsx": "4.19.3",
Error: Process completed with exit code 1.
```

Still haven't ruled out user error (i.e. it could very much be my fault). But wanted to put this out there in case more people run into this

My guess would be on https://github.com/keithamus/sort-package-json/pull/358 by @karlhorky 
Note: I'm actually running prettier-plugin-packagejson, which in turn uses sort-package-json

```sh
❯ node --version && yarn --version
v20.19.2
4.6.0
```

Tobbe:
Different package managers sort dependencies using different sorting orders. This PR tries to detect the package manager that's used by the project, and chooses the sorting order based on the detected package manager

Fixes #363

### DIFF
--- a/index.js
+++ b/index.js
@@ -192,13 +192,11 @@ const sortDependencies = onObject((dependencies, packageJson) => {
  *
  * @see https://docs.npmjs.com/cli/v7/using-npm/workspaces?v=true#running-commands-in-the-context-of-workspaces
  */
-const sortWorkspaces = onObject(
-  pipe([
-    sortObjectBy(['packages', 'catalog']),
-    overProperty('packages', uniqAndSortArray),
-    overProperty('catalog', sortDependencies),
-  ]),
-)
+const sortWorkspaces = pipe([
+  sortObjectBy(['packages', 'catalog']),
+  overProperty('packages', uniqAndSortArray),
+  overProperty('catalog', sortDependencies),
+])
 
 // https://github.com/eslint/eslint/blob/acc0e47572a9390292b4e313b4a4bf360d236358/conf/config-schema.js
 const eslintBaseConfigProperties = [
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -357,153 +394,167 @@ const sortExports = onObject((exports) => {
...
+ *
+ * @param {'npm'|'yarn'|'pnpm'} packageManager - The package manager to use
+ * @returns {Array} fields configuration array
+ */
+function createFields(packageManager) {
+  // npm uses locale-aware comparison, yarn and pnpm use simple string comparison
+  const sortDependencies =
+    packageManager === 'npm'
+      ? sortObjectAlphabeticallyWithLocale
+      : sortObjectAlphabetically
```
keithamus: This might be simpler if you created a `sortObjectAlphabeticallyDetectingPackageManager` or something - making the detection just-in-time, rather than doing so ahead of time. I could imagine it being a smaller diff. I'm not precious about either approach though, so this LGTM.

index.js
```diff
@@ -115,9 +116,55 @@ const sortObjectByIdent = (a, b) => {
...
+
+/**
+ * Sort dependencies alphabetically, detecting package manager to use
+ * appropriate comparison npm uses locale-aware comparison, yarn and pnpm use
+ * simple string comparison
+ *
+ * @param {object} packageJson - The full package.json object for detection
+ * @returns {function} - Sort function for dependencies
+ */
+function sortObjectLikePackageManager(packageJson) {
```
fisker:
```suggestion
function sortDependencies(packageJson) {
```

index.js
```diff
@@ -115,9 +116,55 @@ const sortObjectByIdent = (a, b) => {
...
+  if (json.pnpm) return 'pnpm'
+  if (json.engines?.npm) return 'npm'
+
+  try {
+    if (fs.existsSync('yarn.lock')) return 'yarn'
+    if (fs.existsSync('pnpm-lock.yaml')) return 'pnpm'
+    if (fs.existsSync('package-lock.json')) return 'npm'
+  } catch {
+    // If fs operations fail, default to npm
+  }
```
fisker: This result should be cached.

index.js
```diff
@@ -139,9 +193,12 @@ const sortWorkspaces = (workspaces) => {
...
   }
 
-  // Then add catalog if it exists and sort it like dependencies
+  // Then add catalog if it exists and sort it using locale-aware comparison
+  // (npm-style)
   if (workspaces.catalog) {
-    sortedWorkspaces.catalog = sortDependenciesLikeNpm(workspaces.catalog)
+    sortedWorkspaces.catalog = sortObjectBy((a, b) => a.localeCompare(b, 'en'))(
+      workspaces.catalog,
+    )
```
fisker: I forgot what's this, should it alway sort like NPM? If so, please restore to use `sortDependenciesLikeNpm`.

keithamus: Workspaces are also supported by yarn IIRC, so this should probably use `sortDependencies` also.

https://github.com/keithamus/sort-package-json/blob/main/tests/workspaces.js

> ```js
>       catalog: {
>         'is-even': 'npm:is-even@1.0.0',
>         'is-odd': 'npm:is-odd@3.0.1',
>         'alpha-lib': 'npm:alpha-lib@1.0.0',
>       },
> ```

index.js
```diff
@@ -60,12 +61,12 @@ const sortDirectories = sortObjectBy([
...
-const overProperty =
-  (property, over) =>
-  (object, ...args) =>
+const overProperty = (property, over) =>
+  onObject((object, ...args) =>
     Object.hasOwn(object, property)
       ? { ...object, [property]: over(object[property], ...args) }
-      : object
+      : object,
+  )
```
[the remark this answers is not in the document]
keithamus: Nice find! We can keep it here.

tests/dependency-sorting.js
```diff
@@ -0,0 +1,155 @@
...
+  )
+
+  t.deepEqual(
+    getDependencyOrders({ engine: { npm: '>1.0.0' } }, () => false),
+    npmSortedResult,
+  )
+
+  // packageManager
+  t.deepEqual(
+    getDependencyOrders({ packageManager: 'npm@1.0.0' }, () => {
```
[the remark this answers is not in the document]
keithamus: Should we block on this?

Tobbe:
@keithamus and @fisker Thanks for guiding me through this 🙏

PR #390, opened 2026-01-14
feat: allows `@` character appear before `-`

sidferreira:
We faced this recently when sorting  some dependencies where we wanted
```
react-native@1.2.3
react-native-reanimated@1.2.3
```
instead of
```
react-native-reanimated@1.2.3
react-native@1.2.3
```

As `react-native` is a base for `react-native-*`, I think this is a nice improvement.

### DIFF
--- a/index.js
+++ b/index.js
@@ -166,9 +166,38 @@ function shouldSortDependenciesLikeNpm(packageJson) {
 }
 
 /**
+ * Custom comparator that sorts @ before - when comparing strings.
+ * This ensures that "package@..." comes before "package-..." in dependencies.
+ *
+ * Strategy: Check character by character until we find a difference.
+ * If the difference is @ vs -, handle specially. Otherwise, delegate to
+ * the provided comparison function.
+ */
+const compareWithAtBeforeDash = (a, b, fallbackCompareFn) => {
+  const len = Math.min(a.length, b.length)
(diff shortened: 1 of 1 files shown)


keithamus:
Thanks for this PR!

We'll need to ensure there are new tests added for this, and especially to test the various combinations of dashes and not-dashes. It might be worth collecting a set of package.jsons in the wild to see how this PR changes the sort order for those.

fisker:
> it also adds the version of the patches dependency to the key: react-native@^0.79.5.

We already have the correct sort method for it, it's here https://github.com/keithamus/sort-package-json/blob/e3f2370982fdd22fbf6d68ab8c9690f3a767f834/index.js#L294

sidferreira:
Anyways, thanks for your time reviewing my pr.

fisker:
> Pnpm just adds items to the end of the list, no sorting done in my experience.

If it's true, you can implement it your way you want it be, but don't touch the sort logic for `depdendecies`.

fisker:
Ah, I misunderstood.

If you want 


```
  pnpm
    patchedDependencies
      "react-native@^0.79.5": "...",
      "react-native-appsflyer@^6.15.1": "...",
```

I think `sortObjectBySemver` is suitable.

PR #391, opened 2026-01-14
chore: sort peerDependenciesMeta with depth = 2

hyperz111:
(I add TODO comment to `cli.js` for in future)

Fixture:
```js
const fixture = {
  peerDependenciesMeta: {
    vue: {
      optional: true,
    },
    react: {
      sub: {
        what: false,
        fine: true
      },
      optional: true,
    }
  }
}
```

Before:
```json
{
  "peerDependenciesMeta": {
    "react": {
      "optional": true,
      "sub": {
        "fine": true,
        "what": false
      }
    },
    "vue": {
      "optional": true
    }
  }
}
```

After:
```json
{
  "peerDependenciesMeta": {
    "react": {
      "optional": true,
      "sub": {
        "what": false,
        "fine": true
      }
    },
    "vue": {
      "optional": true
    }
  }
}
```

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,6 +2,7 @@
 import fs from 'node:fs/promises'
 import streamConsumers from 'node:stream/consumers'
 import { parseArgs } from 'node:util'
+// TODO: Use fs.promises.glob when we drop support for Node.js < 22
 import { glob } from 'tinyglobby'
 import sortPackageJson from './index.js'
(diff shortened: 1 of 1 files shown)


hyperz111:
> What's the motivation for this change? If we're sorting, why don't we want to sort deeper than 2 levels?

I see **TODO** comment. But if this isn't needed, i will close this.

PR #392, opened 2026-01-18
chore: exclude git-hooks-list

hyperz111:
`git-hooks-list` is `require()`-able, so i exclude it from CommonJS bundle. From `22.7kb` to `22.1kb`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "index.cjs"
   ],
   "scripts": {
-    "build": "esbuild index.js --bundle --platform=node --outfile=index.cjs --external:semver",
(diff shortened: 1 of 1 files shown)


fisker (approved):

hyperz111:
I think if we should bundle dependencies for our CLI, agree?

hyperz111:
> I guess I'm not sure of the justification for this change. It makes the bundle smaller but that means it's requiring other files right? Are we trading off smaller bundles for longer execution time? Should we unbundle all dependencies? Why just this one?

Maybe you're right. Excluding some dependencies can make the cold start slower. Because the module must load the excluded module.

Before: 102.029ms
After: 144.09ms

PR #411, opened 2026-04-24
feat!: replace tinyglobby with built-in fs.glob

hyperz111:
This PR will replace [`tinyglobby`](https://npmx.dev/package/tinyglobby) with built-in [`fs.glob`](https://nodejs.org/docs/latest/api/fs.html#fspromisesglobpattern-options), but we need to only support Node.js `22.17` or higher.


cli.js
```diff
@@ -71,7 +70,11 @@ async function sortPackageJsonFile(file, reporter, isCheck) {
 }
 
 async function sortPackageJsonFiles(patterns, { ignore, ...options }) {
-  const files = await glob(patterns, { ignore })
+  const results = await Array.fromAsync(fs.glob(patterns, { exclude: ignore }))
+  const files =
+    process.platform === 'win32'
+      ? results.map((result) => result.replaceAll('\\', '/'))
+      : results
```
keithamus, line 77: This doesn't look right. This should use `path.normalize` or so, right?

hyperz111: I think this is good, because `glob` will return array of paths using OS-spesific slash and `path.normalize` doesn't work. I replace slash for keep consistent with snapshots.

fisker (approved):

